PR #25, opened 2016-02-11
Add window watcher code & call placeholder function on window load

### LINKED ISSUE #3: create XBL to insert our hbox
- create XBL file
- grab anonymous hbox el inside bootstrap

### LINKED ISSUE #4: add window watcher code
great example code:
- https://github.com/Mardak/restartless/blob/watchWindows/bootstrap.js
- the old addon code didn't check that the window was loaded before setting a load handler, so, it's not the perfect example, but otherwise it does work

the code should:
- iterate DOM windows and load code into any that exist
- set a window watcher listener for new window creation
- on window created, if it's loaded, load code into window; if not, set a load watcher that loads code
- code loader should load our JS into the app object (not the window global), and should also load our stylesheet into the window

jaredhirsch:
@chuckharmston R? This might make more sense together with the XBL code for issue 3, since the window iteration currently results in nothing happening (though you could set a console.log call inside the `loadIntoWindow` function to convince yourself it's actually getting called). I'm happy to add the XBL code to this PR tomorrow morning, but ran out of time for tonight.

Couple of TODOs in the window watcher code that I'd like to fix before landing, or could just fix in a follow-up PR.

Comments/questions welcome :-)

Adding "fixes #4" to try to help waffle, which seems to be confused about which issue this addresses.

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -2,11 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const Cu = Components.utils;
+const {utils: Cu} = Components;
 
 Cu.import('resource://gre/modules/XPCOMUtils.jsm');
 XPCOMUtils.defineLazyModuleGetter(this, 'UniversalSearch',
-  'chrome://usearch-lib/universal-search.js');
+  'chrome://usearch-lib/content/universal-search.js');
 
 function startup(data, reason) {
   UniversalSearch.load();  
--- a/lib/universal-search.js
+++ b/lib/universal-search.js
@@ -2,13 +2,34 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const {utils: Cu} = Components;
+
+Cu.import('resource://gre/modules/XPCOMUtils.jsm');
+XPCOMUtils.defineLazyModuleGetter(this, 'console',
+  'resource://gre/modules/Console.jsm');
+XPCOMUtils.defineLazyModuleGetter(this, 'WindowWatcher',
+  'chrome://usearch-lib/content/window-watcher.js');
+
 const EXPORTED_SYMBOLS = ['UniversalSearch'];
 
 function Search() {}
 
 Search.prototype = {
-  load: function() {},
-  unload: function() {}
+  load: function() {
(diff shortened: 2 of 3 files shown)


bootstrap.js
```diff
@@ -2,11 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const Cu = Components.utils;
+const {utils: Cu} = Components;
 
 Cu.import('resource://gre/modules/XPCOMUtils.jsm');
 XPCOMUtils.defineLazyModuleGetter(this, 'UniversalSearch',
-  'chrome://usearch-lib/universal-search.js');
+  'chrome://usearch-lib/content/universal-search.js');
```
jaredhirsch, line 9: bugfix from previous broken commit, happy to extract into a separate bugfix commit ¯_(ツ)_/¯

chuckharmston:
screen shot 2016-02-11 at 8 17 03 am

> Adding "fixes #4" to try to help waffle, which seems to be confused about which issue this addresses.

Seems fine. I've found that it likes it better on the first line, though.

bootstrap.js
chuckharmston, line 9: You have to backtick your `¯\_(ツ)_/¯`s, because GitHub wants to make the right arm an escape sequence.

Rookie.

lib/universal-search.js
```diff
@@ -2,13 +2,30 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const {utils: Cu} = Components;
```
chuckharmston, line 5: ES6 :heart_eyes:

lib/window-watcher.js
```diff
@@ -0,0 +1,81 @@
...
+  _onWindowLoaded: function(e) {
+    // TODO: cleaner way to get window pointer?
+    const win = e.target.ownerGlobal;
+    win.removeEventListener('load', ww._onWindowLoaded, false);
+    // TODO: explain why we're checking this
+    if (win.location.href == 'chrome://browser/content/browser.xul') {
+      ww._loadCallback.call(null, win);
+    }
+  },
+  _onError: function(msg) {
```
chuckharmston, line 87: Kind of a style nit, but I love my vertical whitespace. Maybe add a line between each method?

chuckharmston:
This looks great. Feel free to merge as makes sense to you.

lib/window-watcher.js
jaredhirsch, line 87: Sure, sounds good

PR #28, opened 2016-02-16
Adds small library to fetch recommendations (closes #6).

### LINKED ISSUE #6: create DOM-free recommendation server lib
- when a function is called with a string, fire xhr at the rec server
- ensure we have a single xhr at any time; cancel on subsequent requests
- we'll inject one per window, like with other code
- do not use DOM code / window APIs if at all possible (get xhr from the service loader)

chuckharmston:
r? @6a68

### DIFF
--- a/lib/recommendation.js
+++ b/lib/recommendation.js
@@ -1,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const { classes: Cc } = Components;
+const xhr = Cc['@mozilla.org/xmlextras/xmlhttprequest;1'];
+const EXPORTED_SYMBOLS = ['RecommendationServer']; // eslint-disable-line no-unused-vars
+
+
+function RecommendationServer(server, timeout=1000) {
+  /*
+  Class providing a connection to a recommendation server and a method for
+  querying it. This class should be initialized once and exported as a
+  singleton to modules that will use it.
+
+  Usage:
+
+  let recommendation = new RecommendationServer('http://universal-search.dev');
+  */
+  this.server = server;
+  this.request = xhr.createInstance();
+  this.request.timeout = timeout;
+  this.request.isInFlight = false;
+}
+
+RecommendationServer.prototype._url = function(query) {
+  return `${this.server}/?q=${encodeURIComponent(query)}`;
+}
+
+RecommendationServer.prototype.search = function(query) {
+  /*
+  Method to query a recommendation server. Returns a promise that resolves to
+  a recommendation or rejects to a reason that a recommendation is not
+  available.
+
+  Usage:
+
+  recommendation.search('hello wor').then(data => {
+    // Recommendation returned as `data` object literal.
+  }, reason => {
+    // No recommendation returned for `reason` string.
+  });
+  */
+  const endpoint = this._url(query);
+  const req = this.request;
+  return new Promise((resolve, reject) => {
+    try {
+      if (req.isInFlight) {
+        req.abort();
+      }
+      req.addEventListener('abort', evt => {
+        req.isInFlight = false;
+        reject(`Aborted request for "${query}".`);
+      });
+      req.addEventListener('error', evt => {
+        req.isInFlight = false;
+        reject(`Request for "${query}" errored.`);
+      });
+      req.addEventListener('load', evt => {
+        req.isInFlight = false;
+        let recommendation;
+        try {
+          recommendation = JSON.parse(req.responseText);
+        } catch (err) {
+          reject(`Unable to parse JSON for "${query}".`);
+        }
+        if (req.status === 200 && recommendation !== {}) {
+          resolve(recommendation);
+        }
+        reject(`No recommendation available for "${query}".`);
+      });
+      req.open('GET', endpoint);
+      req.isInFlight = true;
+      req.send();
+    } catch (err) {
+      reject(`Uncaught exception in search for "${query}".`);
+    }
+  });
+}


lib/recommendation.js
```diff
@@ -0,0 +1,61 @@
+var XMLHttpRequest = Cc['@mozilla.org/xmlextras/xmlhttprequest;1'].createInstance();
```
jaredhirsch: I've been using `const` in general in this code base, rather than `var`

lib/recommendation.js
```diff
@@ -0,0 +1,61 @@
...
+
+
```
jaredhirsch, line 8: You need EXPORTED_SYMBOLS here, see this MDN page for details: https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Using

lib/recommendation.js
jaredhirsch: Ah, you also need to define `Cc` before you use it. Have a look at other files in this repo to see how.

lib/recommendation.js
```diff
@@ -0,0 +1,61 @@
...
+      // Recommendation returned as `data` object literal.
+    }, reason => {
+      // No recommendation returned for `reason` string.
+    });
+
+    */
+    const url = this._url(query);
+    return new Promise((resolve, reject) => {
+      try {
+        var req = new XMLHttpRequest();
```
jaredhirsch: This library should keep a single xhr pointer and abort in-flight requests when new searches begin

lib/recommendation.js
```diff
@@ -0,0 +1,61 @@
...
+  */
+  constructor(server) {
+    this.server = server;
+  }
+
+  _url(query) {
+    return `${this.server}/?q=${encodeURIComponent(query)}`;
+  }
+
+  make(query) {
```
jaredhirsch: `make` seems like an odd API choice. How about `search`?

lib/recommendation.js
```diff
@@ -0,0 +1,61 @@
+var XMLHttpRequest = Cc['@mozilla.org/xmlextras/xmlhttprequest;1'].createInstance();
+
+
+class RecommendationServer {
```
jaredhirsch: I'm using prototypal inheritance elsewhere, let's be consistent and use that for now

lib/recommendation.js
```diff
@@ -0,0 +1,61 @@
...
+        var req = new XMLHttpRequest();
+        req.addEventListener('abort', (evt) => {
+          reject('Aborted request.');
+        });
+        req.addEventListener('error', (evt) => {
+          reject('Errored request.');
+        });
+        req.addEventListener('load', (evt) => {
+          if (req.status === 200 && req.responseText !== {}) {
+            resolve(JSON.parse(req.responseText));
```
jaredhirsch: JSON calls should be wrapped in a try-catch, else the promise may explode and never resolve

lib/recommendation.js
jaredhirsch: Note also that FF only enabled `class` in FF 45 ([mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes), [tracking bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1197932)), and 45 doesn't go GA until 3/8/15. It might get backed out, and I'm not sure about adding a transpiler to this repo (though we could talk about it in a discussion bug, if you want to file one).

lib/recommendation.js
```diff
@@ -0,0 +1,61 @@
...
+
+  let recommendation = new RecommendationServer('http://universal-search.dev');
+
+  */
+  constructor(server) {
+    this.server = server;
+  }
+
+  _url(query) {
+    return `${this.server}/?q=${encodeURIComponent(query)}`;
```
jaredhirsch: :+1: nice use of template strings

lib/recommendation.js
jaredhirsch: Alas, one more line note here: we should lazily instantiate the xhr, because FF has problems with slow initial loading

jaredhirsch:
@chuckharmston I've tried to cover most of the issues in the first pass. Ping me if you have any questions, happy to help

lib/recommendation.js
```diff
@@ -0,0 +1,61 @@
...
+          reject('Errored request.');
+        });
+        req.addEventListener('load', (evt) => {
+          if (req.status === 200 && req.responseText !== {}) {
+            resolve(JSON.parse(req.responseText));
+          }
+          reject('No recommendation available.');
+        });
+        req.open('GET', url);
+        req.send();
```
jaredhirsch: Might be good to set a timeout before sending the request? That could also be specified in the constructor.

jaredhirsch:
@chuckharmston This review came out a bit more terse than usual, please don't hesitate to ask if anything's unclear or you'd like suggestions

lib/recommendation.js
chuckharmston: You don't _search_ recommendations, you _make_ them. But that's just my twisted logic; happy to change :)

chuckharmston:
@6a68: this addresses everything <s>except for the prototypical inheritance bit, which I'll hit next</s>. The test script I've been using:

``` javascript
var recommendation = new RecommendationServer('http://universal-search.dev');
Components.utils.import("resource://gre/modules/Console.jsm");

const rec = query => {
  recommendation.search(query).then(data => {
    console.log('Resolved', data)
  }, reason => {
    console.log('Rejected', reason)
  });
}

rec('m');
rec('ma');
rec('mar');
rec('mart');
setTimeout(() => {
  rec('mart');
}, 500);
```

And the output when ran:

```
Rejected Aborted request for "m".
Rejected Aborted request for "ma".
Rejected Aborted request for "mar".
Rejected No recommendation available for "mart".
// 500ms delay...
Resolved Object { enhancements: Object, query: Object, result: Object }
```

lib/recommendation.js
jaredhirsch: lol

lib/recommendation.js
```diff
@@ -0,0 +1,77 @@
...
+      if (req.inFlight) {
+        req.abort();
+      }
+      req.addEventListener('abort', (evt) => {
+        req.inFlight = false;
+        reject(`Aborted request for "${query}".`);
+      });
+      req.addEventListener('error', (evt) => {
+        req.inFlight = false;
+        reject(`Aborted request for "${query}".`);
```
chuckharmston: Copypasta

jaredhirsch:
@chuckharmston Nice work! Seriously, this looks really great.

It's unfortunate that nobody writes blog posts about xhr or xhr2 anymore, because the API has gotten pretty large, and there's not many helpful resources aside from the [whatwg spec](https://xhr.spec.whatwg.org/). So, I don't have great references to share, but I think the following should work and be simpler:

```
RecommendationServer() {
  // ... other stuff ...
  this.xhr = null;
}
RecommendationServer.prototype.search = function(query) {
  // cancel anything that's in flight
  if (this.xhr) {
    this.xhr.abort(); // this will auto-reject the previous promise
  }
  this.xhr = xhr.createInstance();
  this.xhr.timeout = timeout;
  this.xhr.onloadend = () => {
    if (this.xhr.status == 200) {
      // success! parse this.xhr.responseText and resolve()
    }
    // something went wrong, reject()
  }
  this.xhr.open('GET', endpoint);
  this.xhr.send();
};
```

I think you only need to wrap the `xhr.open` (in case the server URL is typo'd) and the `JSON.parse` in a try/catch.

One other nit: how about `RecommendationService` instead of `RecommendationServer`?

jaredhirsch:
Oh, in my response above I forgot to wrap the contents of the search function in a `return new Promise`, but I think it should otherwise be accurate

chuckharmston:
I actually intentionally avoided using `loadend`; there are qualitative differences between a request rejected because it was aborted (desired and good), a request rejected because of a server error (400 or 500), one rejected because a recommendation wasn't available (i.e. a 202), and one rejected for invalid JSON. We'd at least want to log those differently, right?

chuckharmston:
Also: I was reusing XHR objects to avoid the memory overhead of garbage cleaning completed XHR instances and instantiating new ones for every keystroke. Your proposal is a little simpler, but in my estimation we'd prefer performance to simplicity.

lib/recommendation.js
```diff
@@ -0,0 +1,77 @@
...
+        req.isInFlight = false;
+        reject(`Aborted request for "${query}".`);
+      });
+      req.addEventListener('error', evt => {
+        req.isInFlight = false;
+        reject(`Request for "${query}" errored.`);
+      });
+      req.addEventListener('load', evt => {
+        req.isInFlight = false;
+        if (req.status === 200 && req.responseText !== {}) {
```
jaredhirsch: Should this be `req.responseText !== '{}'`, since the response hasn't yet been parsed?

jaredhirsch:
One quick question, otherwise I'll verify it works tomorrow early and land it :beers:

lib/recommendation.js
chuckharmston: Yeah, it definitely should be. Great catch.

chuckharmston:
I updated it to parse the JSON before doing anything, rather than doing a string check.

lib/recommendation.js
```diff
@@ -0,0 +1,78 @@
...
+        reject(`Aborted request for "${query}".`);
+      });
+      req.addEventListener('error', evt => {
+        req.isInFlight = false;
+        reject(`Request for "${query}" errored.`);
+      });
+      req.addEventListener('load', evt => {
+        req.isInFlight = false;
+        try {
+          const recommendation = JSON.parse(req.responseText);
```
jaredhirsch: Sadly, `const` is block-scoped, so this needs to be

```
let recommendation;
try {
  recommendation = JSON.parse(...)
}
```

jaredhirsch:
@chuckharmston One minor block-scoping thing, feel free to merge once that's resolved

chuckharmston:
Thanks!

Issue #30, opened 2016-02-17
Ensure XBL works in Nightly


jaredhirsch:
I'm intermittently not seeing our XBL applied in Nightly. In these cases, Nightly seems to crash within 30 sec of startup. Ugh.

jaredhirsch:
Verified that #31 works in all channels. Closing :beers:

Issue #7, opened 2016-02-10
possibly remove search box?


jaredhirsch:
when the telemetry experiment around search box is done, possibly revisit this
- use the customize scripting code from the addon to hide the search box if it's shown, and on uninstall, if it was shown, put it back

PR #31, opened 2016-02-17
Insert hbox by overriding XBL bindings of existing popup

### LINKED ISSUE #3: create XBL to insert our hbox

jaredhirsch:
@chuckharmston R?

This patch uses selector specificity to insert our hbox into the existing popup. You can tell it's attached by inspecting the popup in the browser toolbox: it's under the `popupset` with id `mainPopupSet`, and our popup is the `panel` with id `PopupAutoCompleteRichResult`. There should be an hbox inside the popup with the anonid 'recommendation'. Here's a screenshot of how it looks in release FF:

screen shot 2016-02-17 at 9 11 55 am

I'm having some problems with Nightly. It seems like the XBL change is causing Nightly to fail to apply our styles, and then crash. Not sure if this is our changes or something busted in Nightly. I've opened #30 to investigate.

---

Original commit message:

Rather than inserting a new popup, we can use CSS specificity to
override the existing popup's XBL bindings, inserting our hbox above the
list of results.

Holding off on adding UI classes in this commit, so we still aren't
injecting any scripts into each window.

Closes #3.

### DIFF
--- a/chrome.manifest
+++ b/chrome.manifest
@@ -1,1 +1,3 @@
-content usearch-lib lib/
+content universalsearch chrome/content/
+content universalsearch-lib lib/
+content universalsearch-skin chrome/skin/
--- a/chrome/skin/style.css
+++ b/chrome/skin/style.css
@@ -1,0 +1,10 @@
+/* Use extreme selector specificity to override the default bindings,
+ * defined in chrome://browser/content/browser.css, which use the single
+ * id selector #PopupAutoCompleteRichResult.
+ */
+popupset#mainPopupSet panel#PopupAutoCompleteRichResult {
+  /* This odd CSS property connects an XBL file with XUL DOM.
+   * See mdn.io/moz-binding for more.
+   */
+  -moz-binding: url("chrome://universalsearch/content/binding.xml#recommendation-popup");
+}
--- a/lib/universal-search.js
+++ b/lib/universal-search.js
@@ -10,7 +10,7 @@ XPCOMUtils.defineLazyModuleGetter(this, 'console',
 XPCOMUtils.defineLazyModuleGetter(this, 'CustomizableUI',
   'resource:///modules/CustomizableUI.jsm');
 XPCOMUtils.defineLazyModuleGetter(this, 'WindowWatcher',
-  'chrome://usearch-lib/content/window-watcher.js');
+  'chrome://universalsearch-lib/content/window-watcher.js');
 
 const EXPORTED_SYMBOLS = ['UniversalSearch'];
 
@@ -18,7 +18,7 @@ function Search() {}
 
 Search.prototype = {
   load: function() {
-    WindowWatcher.start(this.loadIntoWindow, this.unloadFromWindow, this.onError);
+    WindowWatcher.start(this.loadIntoWindow, this.unloadFromWindow, this.onError, this);
   },
 
   unload: function() {
@@ -27,17 +27,14 @@ Search.prototype = {
 
   loadIntoWindow: function(win) {
     console.log('loadIntoWindow start');
-    const document = win.document;
 
     // Sets the app global per window.
     win.universalSearch = win.universalSearch || {}
 
-    // If the search bar is visible, note the location and hide it.
-    const searchBarLocation = CustomizableUI.getPlacementOfWidget('search-container');
-    if (searchBarLocation) {
-      win.universalSearch.searchBarLocation = searchBarLocation;
-      CustomizableUI.removeWidgetFromArea('search-container');
-    }
+    this._hideSearchBar(win);
+    this._loadStyleSheets(win);
+    this._loadScripts(win);
+    this._startApp(win);
 
     console.log('loadIntoWindow finish');
   },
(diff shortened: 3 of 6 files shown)


chrome.manifest
```diff
@@ -1 +1,2 @@
+content usearch chrome/
```
chuckharmston: Is it possible to use better nomenclature for this? `universalsearch` would be so much clearer.

chrome/skin/style.css
```diff
@@ -0,0 +1,10 @@
+/* Use extreme selector specificity to override the default bindings,
+ * defined in chrome://browser/content/browser.css, which use the single
+ * id selector #PopupAutoCompleteRichResult.
+ */
+popupset#mainPopupSet panel#PopupAutoCompleteRichResult {
+  /* This odd CSS property connects an XBL file with XUL DOM.
+   * See mdn.io/moz-binding for more.
+   */
+  -moz-binding: url("chrome://usearch/content/content/content.xml#usearch-urlbar-rich-result-popup");
```
chuckharmston: Would `!important` accomplish the same thing? If so, would it be cleaner?

chuckharmston:
A bit of a debugging dump:
- With the add-on enabled, the popup never opens.
- With the add-on disabled, the popup does open.
- The CSS is always inserted into the DOM.
- Sometimes the correct `-moz-binding` is applied, but sometimes it uses the default. This is intermittent on browser restart. I don't see any pattern in this.
- Neither [chrome://usearch/content/content/content.xml](chrome://usearch/content/content/content.xml) nor [chrome://usearch/content/content/content.xml#usearch-urlbar-rich-result-popup](chrome://usearch/content/content/content.xml#usearch-urlbar-rich-result-popup) load in the browser.

Happy to help debug how I'm able!

jaredhirsch:
@chuckharmston OK, that's interesting and typically (for XBL) weird.

Could you try with a fresh profile? Maybe that'll help.

I'm not sure why you wouldn't be able to surf to the content.xml file. If the stylesheet is applied to the popup, but the xml file isn't found, then I think it makes sense for the popup to be broken.

Does the chrome.manifest look the same in your clone as it does in the branch on github?

jaredhirsch:
@chuckharmston Here are some more ideas for debugging, just going down the list:
- start with a new, empty profile
- open `about:config`, set `xpinstall.signatures.required` to `false`
  - release channel FF seems to reset this to `true`, which leads to the proxy file getting silently deleted. (pretty annoying, I wonder if there's a workaround, maybe a user.js file?)
- create a proxy file in the new, empty profile (you'll have to create the extensions directory)
  - the name should be `universal-search@mozilla.com`
  - the contents should be the absolute path to the repo, with a trailing slash
- next, open FF using the new profile
- be sure you see the universal search addon in the about:addons list
- try to load the XBL chrome URL: `chrome://usearch/content/content/content.xml` (I can do something about that name, it's not great)
- see if you can load the stylesheet using the chrome URL `chrome://usearch/content/skin/style.css`

Hopefully it's one of these relatively simple things. If not, I guess we'll keep digging :-)

chuckharmston:
After quite a bit more poking*, I was able to find the `<hbox>` in the DOM and load both of the XBL chrome file and the stylesheet. However, the popup still doesn't appear when the addon is enabled. If that's expected, merge it! If not, let me know and we can keep poking to figure out why I can't reproduce.

\* Some of the devprefs settings (including `xpinstall.signatures.required`)  were being unset on start. As a result, the proxy was being deleted on startup in release, but the addon was still listed in about:addons (‽‽).

jaredhirsch:
> After quite a bit more poking*, I was able to find the <hbox> in the DOM and load both of the XBL chrome file and the stylesheet.

Yay! Welcome to the funhouse mirror version of web dev that is Gecko hacking...

> However, the popup still doesn't appear when the addon is enabled. If that's expected, merge it! If not, let me know and we can keep poking to figure out why I can't reproduce.

Hmm, the popup should definitely be shown--especially if you see the node in the DOM and the stylesheet's in the DOM, too. Do you see any errors in the browser toolbox (that aren't also visible with the addon disabled)?

Sometimes, if the popup's just gone, it means the XBL is silently failing (XBL seems to swallow exceptions, adding to the "fun" of debugging). However, in that case, I should be seeing it as well in a fresh profile.

Another idea: you can clear the browser toolbox console, then open a new window (which will do all the popup modification), and look for errors. Because the enable/disable process adds/removes our stylesheet from the XUL DOM, you could also try clearing the console, then disable and re-enable the add-on, and see what happens to the popup node / in the console.

> - Some of the devprefs settings (including xpinstall.signatures.required) were being unset on start. As a result, the proxy was being deleted on startup in release, but the addon was still listed in about:addons (‽‽).

Yeah, what happens is: the pref gets auto-flipped; the addon gets set to disabled; if you don't flip the signature pref before closing firefox, the proxy file will get auto-deleted.

chrome.manifest
jaredhirsch: :+1: I'll change that. I can also change the mapping here and the xml file name to avoid "content/content/content.xml" in the chrome URL, which seems suboptimal

chuckharmston: theresnoplacelikehome/theresnoplacelikehome/theresnoplacelikehome.xml

jaredhirsch:
@chuckharmston mind giving this another look? Seems to be reliably working in both Nightly and Release channels; thanks to using an `id` instead of an `anonid`, you can easily verify it works: `document.getElementById('universal-search-recommendation')`.

I'm happy to rebase this after your search bar customization PR lands, feel free to merge that first.

jaredhirsch:
There was a ton of XBL discussion on IRC earlier today. The end result here is that I copied over [all of the `<content>`](https://dxr.mozilla.org/mozilla-central/source/browser/base/content/urlbarBindings.xml#1155-1190) of the original popup, except for the "do you want to enable search suggestions" node. I removed that node because I was getting weird l10n-related errors, but, now that I think about it, we probably need to set the search suggestion preference to true. Filed as #34.

chuckharmston:
This is working flawlessly for me. Very nicely done! #7 has been merged, feel free to rebase and merge this.

lib/universal-search.js
```diff
@@ -27,34 +27,88 @@ Search.prototype = {
 
   loadIntoWindow: function(win) {
     console.log('loadIntoWindow start');
-    const document = win.document;
 
     // Sets the app global per window.
     win.universalSearch = win.universalSearch || {}
```
jaredhirsch, line 32: I was on the fence about pulling this into a method, just to make the loadIntoWindow more uniform, hence, more scannable, but then, it's a single line of code. Thoughts?

jaredhirsch, line 32: Eh, I'll just land it and we can discuss later :-)

PR #32, opened 2016-02-18
Hides the search box on loadIntoWindow; restores it on unloadFromWindow (closes #7).

### LINKED ISSUE #7: possibly remove search box?

chuckharmston:
r? @6a68 

screen-recording-2016-02-18-13-51-11

### DIFF
--- a/lib/universal-search.js
+++ b/lib/universal-search.js
@@ -7,6 +7,8 @@ const {utils: Cu} = Components;
 Cu.import('resource://gre/modules/XPCOMUtils.jsm');
 XPCOMUtils.defineLazyModuleGetter(this, 'console',
   'resource://gre/modules/Console.jsm');
+XPCOMUtils.defineLazyModuleGetter(this, 'CustomizableUI',
+  'resource:///modules/CustomizableUI.jsm');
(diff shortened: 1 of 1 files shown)


lib/universal-search.js
```diff
@@ -23,9 +25,46 @@ Search.prototype = {
...
+    console.log('loadIntoWindow start');
+    const document = win.document;
 
-  unloadFromWindow: function(win) {},
+    // Sets the app global per window.
+    if (win.universalSearch === undefined) {
+      Object.defineProperty(win, 'universalSearch', {
+        configurable: true,
+        value: {}
+      });
```
jaredhirsch: I think it's actually good enough to do `win.universalSearch = win.universalSearch || {}`. I was looking at this recently and I don't have notes on what motivated this approach in the old popup--maybe some long-forgotten Gecko copypasta.

jaredhirsch:
@chuckharmston Looks great! Aside from getting rid of the Object.defineProperty approach (unless you know of a good reason to prefer it over the "foo = foo || {}" idiom), I think this is good to land.

PR #37, opened 2016-02-24
WIP: App shell

jaredhirsch:
@chuckharmston R? This isn't quite ready to land yet, due to one annoying bug. Probably easier to go through it commit by commit, rather than read the whole diff.

The remaining problem is that we need to wait until our binding has been applied to the XUL DOM before we can try to load the ui code, because that code assumes the recommendation node is in the DOM. I tried waiting for the stylesheet to fire its `load` event, but it turns out that's not good enough: you need to fire a custom DOM event from the binding's `<constructor>` element/method. I'm out of time for today, so opening this as a WIP for now. I might wind up just polling for the element if firing an event gets too weird; in either case, it should hopefully be working early tomorrow.


README.md
```diff
@@ -32,3 +32,14 @@ By participating in this project, you agree to abide by our [code of conduct](./
...
+
+## Coding conventions
+
+- lib/universal-search.js loads code into windows and constructs the object graph
+- use constructors to assign dependencies to the object, and bind `this` to callbacks
+- use `init` for initialization work other than dependency injection
+- use `destroy` to unset anything that could leak memory
+- leading underscore in prototype methods indicates private API
+- event naming convention: hyphenated lower case.
+	- we don't need to namespace it to individual items because it's such a small list of events. just keep it simple.
```
chuckharmston, line 44: This is fantastic, and exactly the sort of thing that will make my life easier. Thanks so much for this.

lib/ui/popup.js
```diff
@@ -0,0 +1,22 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const EXPORTED_SYMBOLS = ['Popup'];
+
+function Popup(opts) {
+  this.win = opts.win;
+  this.events = opts.events;
+
+  this.onPopupHiding = this.onPopupHiding.bind(this);
```
chuckharmston: Nit, but it's sort of a convention to use the infinitive of a verb, rather than a conjugated form, in event names. So `onPopupHide`.

lib/recommendation-server.js
```diff
@@ -0,0 +1,109 @@
...
+  recommendation.init();
+  */
+  this.events = opts.events;
+  this.server = opts.server;
+  this.timeout = opts.timeout;
+  // note: xhr service, not xhr instance
+  this.xhr = opts.xhr;
+  this.console = opts.console;
+
+  this.search = this.search.bind(this);
```
chuckharmston, line 31: Any reason not to run init here? Or just stick the init logic in the constructor?

lib/recommendation-server.js
```diff
@@ -0,0 +1,109 @@
...
+  _url: function(query) {
+    return `${this.server}/?q=${encodeURIComponent(query)}`;
+  },
+  search: function(query) {
+    this._search(query).then(data => {
+      this.events.publish('recommendation', data);
+    }, reason => {
+      // TODO: send this information to metrics instead of logging
+      console.error(reason);
+    });
```
chuckharmston, line 57: I'm not sure this is the correct place for the logic. Wouldn't you want this to be in some sort of controller, rather than in a library that exposes it?

lib/ui/popup.js
jaredhirsch: Yeah, this is a weird one. In this case the name of the event is actually `popuphiding`, hence, `onPopupHiding`: https://developer.mozilla.org/en-US/docs/Web/Events/popuphiding.

The event is sent just before the popup hides itself, so maybe `beforePopupHide` or `beforePopupClose` would be more expressive here?

lib/recommendation-server.js
jaredhirsch, line 57: `¯\_(ツ)_/¯` Good question!

I've aimed for fewer components with more/mixed responsibilities, and this was the simplest way to fire a signal when the response is received. That said, this code is _clearly_ half thought through; I'm totally open to suggestions, but bear in mind our unfortunately typical time crunch.

lib/recommendation-server.js
jaredhirsch, line 31: Like I mentioned in the PR description, I have an (admittedly unlikely) dream that this code might live long enough to justify investing in some unit tests. By separating object graph construction / dependency injection from actual initialization, it's really easy to _just_ construct the object in setup methods, and _just_ exercise the initialization code in individual tests, rather than mixing the two.

The current approach is a little awkward, because we're plucking individual deps off the `opts` object, rather than just assigning `this.app`, but the idea there is to be super explicit about dependencies. Also kinda awkward that we're doing `this` binding in the constructor, but Gecko is prone to memory issues, so we really need detachable callbacks, and this is one reasonably palatable way to get there. If we had another month, I'd like to explore Babel and ES6/7 approaches, but we don't.

Does this help? Are you convinced? Let's talk more if not.

README.md
jaredhirsch, line 44: I'm glad to hear it! I was worried the notes were a little too sloppy and cryptic to be useful in their current form.

jaredhirsch:
@chuckharmston OK, I think I've got the DOM polling working (added a `_pollForElement` method to the recommendation UI class). You can verify it's working by opening the popup once (seems like the element isn't created until the popup is opened once), then doing the following in the browser toolbox:

``` js
window.universalSearch.recommendation.el.collapsed = false;
window.universalSearch.recommendation.el.textContent = "recommend all the things!"
```

Then, open the popup again, and you should see something like:

screen shot 2016-02-24 at 10 34 33 am

lib/ui/popup.js
jaredhirsch: @chuckharmston updated the callback to `beforePopupHide`, and the event name to `before-popup-hide`, thoughts?

chuckharmston:
Brilliant.

screen shot 2016-02-25 at 10 00 53 am

lib/recommendation-server.js
chuckharmston, line 31: From IRC, for posterity:

```
<_6a68> chuck: hey, what do you think of that PR branch? I'm not sure if it's solving the problem of app boilerplate, or really just half-finishing all the things. How's it look to you?
<chuck> I wouldn’t have made all the same architectural decisions, but it’s all academic. If it works, let’s be pragmatic and move on.
```

PR #49, opened 2016-02-29
Enable search suggestions on install/enable; reset on uninstall/disable (closes #34).

### LINKED ISSUE #34: Set search suggestion in awesomebar pref to 'true'
We are not bringing over the "do you want to enable searches?" node from the default popup, because of l10n bugs.

Since the user won't get prompted, we should just set that pref to 'true' on install, if the user hasn't set the pref, and, on uninstall, unset the pref, if the user hadn't set it.

chuckharmston:
r? @6a68

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -5,17 +5,29 @@
 const {utils: Cu} = Components;
 
 Cu.import('resource://gre/modules/XPCOMUtils.jsm');
+XPCOMUtils.defineLazyModuleGetter(this, 'Services',
+  'resource://gre/modules/Services.jsm');
 XPCOMUtils.defineLazyModuleGetter(this, 'UniversalSearch',
   'chrome://universalsearch-lib/content/universal-search.js');
 
 function startup(data, reason) {
-  UniversalSearch.load();  
+  // On enable/install, turn on the search suggestion pref.
+  if (reason === ADDON_ENABLE || reason === ADDON_INSTALL) {
+    Services.prefs.setBoolPref('browser.urlbar.suggest.searches', true);
+  }
+
+  UniversalSearch.load();
 }
 
 function shutdown(data, reason) {
   // Clean up on uninstall or deactivation, but not for normal shutdown.
   if (reason === APP_SHUTDOWN) {
     return;
+  }
+
+  // On disable/uninstall, reset the search suggestion pref.
+  if (reason === ADDON_DISABLE || reason === ADDON_UNINSTALL) {
+    Services.prefs.clearUserPref('browser.urlbar.suggest.searches');
   }
 
   UniversalSearch.unload();


lib/universal-search.js
```diff
@@ -70,6 +74,24 @@ Search.prototype = {
     }
   },
 
+  _enableSearchSuggestions: function(win) {
+    // Enable the browser.search.suggest.enabled pref. Save the previous state
+    // so we can restore it on unload.
+    const searchSuggestIsEnabled = Services.prefs.getBoolPref('browser.search.suggest.enabled');
```
jaredhirsch: I think the pref we want is `browser.urlbar.suggest.searches`. Have a look at [this line](https://github.com/mozilla/gecko-dev/blame/8832806e/testing/profiles/prefs_general.js#L332), seems like that's the pref that was introduced by the UnifiedComplete work.

Because this pref was recently added, it's possible that it won't exist in a given profile. Unfortunately, `getBoolPref` will throw if the pref doesn't exist, so you'll need to use `Services.prefs.getPrefType` to safely check for the pref's existence first.

lib/universal-search.js
```diff
@@ -32,6 +34,7 @@ Search.prototype = {
     win.universalSearch = win.universalSearch || {}
 
     this._hideSearchBar(win);
+    this._enableSearchSuggestions(win);
```
jaredhirsch: Hmm. The code inside `loadIntoWindow` runs in every XUL window, but because prefs are global, and we only need to adjust this pref on install / uninstall, I think we only want to run this code once, as part of the install/uninstall process. I'm not 100% sure about this, but it looks like the about:config page [flushes changes to disk](https://dxr.mozilla.org/mozilla-central/source/mobile/android/chrome/content/config.js#160-161) using `Services.prefs.savePrefFile(null)`. I bet that approach will work here; it seems to be used in some other important-seeming spots when I search for it in dxr.

In order to run on install but not regular startup, the bootstrap.js `startup` method needs to be updated to check that `reason` is ADDON_INSTALL or ADDON_ENABLE (the constants are explained [here](https://developer.mozilla.org/en-US/Add-ons/Bootstrapped_extensions#Reason_constants)), and if it's one of those reasons, call `UniversalSearch.enableSearchSuggestions`.

I think we'll also need to add similar checks in the bootstrap.js `unload` method. Unfortunately, I'm not sure where we can put the user's original preference such that it persists across restarts (I don't think setting it on the window will work, though it seemed to work in the old addon; I could be wrong here). We could either try to figure out how to save this in web storage, or we could just unset the preference entirely, via [`Services.prefs.clearUserPref`](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIPrefBranch#clearUserPref%28%29).

lib/universal-search.js
chuckharmston: Good catch; I didn't realize there was a new pref used for UnifiedComplete. [browser.search.suggest.enabled](http://kb.mozillazine.org/Browser.search.suggest.enabled) has been around for ages but [doesn't appear to be used for anything](https://dxr.mozilla.org/mozilla-central/search?q=browser.search.suggest.enabled&redirect=false&case=false) anymore.

lib/universal-search.js
chuckharmston: Opting for the simple approach here: set it to true on enable/install, reset on disable/uninstall. We can revisit after launch if it becomes a problem.

chuckharmston:
I think this is ready for rereview. Notes:
- I don't think we need to flush prefs to disk. I couldn't create a scenario where it wasn't persisted after restarts.
- It's unfortunate that after a disable, users will need to opt back in to search suggestions. I'm not sure how real of a problem that actually is though.
- We should make sure to explain that this will happen in the Test Pilot experiment detail page.

jaredhirsch:
Looks great! Merging :tada:

PR #51, opened 2016-02-29
Adds library to calculate dominant color of an image from its URL (closes #41).

### LINKED ISSUE #41: add favicon color calculation to recommendation view
Separating this out from the main "render the recommendation" bug (#5)  because it's a bit complex.

See also discussion in #27

chuckharmston:
r? @6a68

### DIFF
--- a/lib/favicon-color.js
+++ b/lib/favicon-color.js
@@ -1,0 +1,90 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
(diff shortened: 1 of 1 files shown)


lib/favicon-color.js
```diff
@@ -0,0 +1,83 @@
+function Favicon(src) {
```
jaredhirsch, line 8: Looks good, but you need some header boilerplate:

``` js
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */

const EXPORTED_SYMBOLS = ['Favicon'];
```

Once that's added, feel free to merge!

PR #52, opened 2016-02-29
Adds preference to set recommendation server (closes #17).

### LINKED ISSUE #17: Allow recommendation server hostname to be set as addon config.

chuckharmston:
r? @6a68 

screen shot 2016-02-29 at 3 12 17 pm

It's primitive, but I'm not sure there are better options, since we're modifying XBL and are thus unable to use the SDK.

Shortcut to get the value of the pref: `window.universalSearch.prefs.getCharPref('server')`

### DIFF
--- a/chrome/content/preferences.xul
+++ b/chrome/content/preferences.xul
@@ -1,0 +1,14 @@
+<?xml version="1.0"?>
+
+<!DOCTYPE mydialog SYSTEM "chrome://myaddon/locale/mydialog.dtd">
+
+<vbox xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul">
+  <setting pref="extensions.universalsearch.server" type="menulist" title="Recommendation Server">
+    <menulist>
+      <menupopup>
+        <menuitem value="http://universal-search.dev" label="http://universal-search.dev"/>
+        <menuitem value="https://search.testpilot.firefox.com" label="https://search.testpilot.firefox.com"/>
+      </menupopup>
+    </menulist>
+  </setting>
+</vbox>
--- a/lib/recommendation-server.js
+++ b/lib/recommendation-server.js
@@ -16,17 +16,17 @@ function RecommendationServer(opts) {
     // events is an instance of lib/events
     events: events,
     xhr: Cc['@mozilla.org/xmlextras/xmlhttprequest;1'],
-    server: 'http://universal-search.dev',
+    prefs: prefs,
     timeout: 1000
   });
   recommendation.init();
   */
   this.events = opts.events;
-  this.server = opts.server;
   this.timeout = opts.timeout;
   // note: xhr service, not xhr instance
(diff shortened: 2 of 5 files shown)


chrome/content/preferences.xul
```diff
@@ -0,0 +1,14 @@
...
+  <prefpane id="unviersalsearch-pane">
+    <preferences>
+      <preference id="pref_server" name="extensions.universalsearch.server" type="string"/>
+    </preferences>
+    <hbox align="center">
+      <label control="server" value="Server: "/>
+      <textbox preference="pref_server" id="server"/>
+    </hbox>
+  </prefpane>
+</prefwindow>
```
jaredhirsch: After some digging in MDN, I think Inline Options are the way to go: that'll give us the controls visible inside about:addons. https://developer.mozilla.org/en-US/Add-ons/Inline_Options

lib/universal-search.js
```diff
@@ -53,6 +54,12 @@ Search.prototype = {
     console.log('unloadFromWindow finish');
   },
 
+  _initializePrefs: function(win) {
+    return Components.classes['@mozilla.org/preferences-service;1']
+      .getService(Components.interfaces.nsIPrefService)
+      .getBranch('extensions.universalsearch.');
```
jaredhirsch: I think `Services.prefs.getBranch` would be simpler here

lib/recommendation-server.js
```diff
@@ -46,7 +46,8 @@ RecommendationServer.prototype = {
     this.events.unsubscribe('printable-key', this.search);
   },
   _url: function(query) {
-    return `${this.server}/?q=${encodeURIComponent(query)}`;
+    const server = this.prefs.getCharPref('server');
```
jaredhirsch, line 50: We should make sure the user hasn't munged the value. One way to do this is to use the `nsIURIFixup` service, which is available on Services.jsm. I _think_ the following should do it:

``` js
// nsIURI instance
const uri = Services.uriFixup(server, Services.uriFixup.FIXUP_FLAG_FIX_SCHEME_TYPOS);
const uriString = uri.asciiSpec;
```

Then pass the uriString into the template string.

chuckharmston, line 50: Switching to inline options, I'm just going to use a `<menulist>` to prevent this.

jaredhirsch:
@chuckharmston Sorry about the prefs panel waste of time, let's just do inline options and address the couple things I mentioned above, and feel free to merge (unless you want another glance before landing).

chrome/content/preferences.xul
chuckharmston: I found some documentation somewhere that said inline options should only be used when you want prefs to be modifiable by touching pieces of chrome that your addon adds—e.g. toolbar buttons. I can't find that though, so I'll push forward.

chuckharmston:
screen shot 2016-03-02 at 10 52 44 am

PR #53, opened 2016-02-29
Override the default urlbar key handler with a no-op handler

### LINKED ISSUE #42: urlbar: add a no-op key handler to XBL
We need to sometimes steal key events from the existing flow.

It looks like the existing code listens for key events in a [truly strange way](https://dxr.mozilla.org/mozilla-central/source/toolkit/content/widgets/autocomplete.xml#617):

```
 <handler event="keypress" phase="capturing"
  action="return this.onKeyPress(event);"/>
```

I think we can optionally handle events by doing something of the form

```
 <handler event="keypress" phase="capturing"
  action="return window.universalSearch.urlbar.onKeyPress(event) || this.onKeyPress(event);"/>
```

If our code handles the key event, and returns true, then the browser should(?) short-circuit the XBL code; otherwise, we can return false, and then the existing XBL handler will get the event.

This seems weird enough to deserve its own issue :tada:

jaredhirsch:
@chuckharmston R?

It turns out that we need to extend the urlbar binding, not the popup binding, to capture keys. Once I realized I was listening to the wrong element, it was pretty easy :-P

You can verify that this is working by replacing the `return this.handleKeyPress` line with something like `alert('keypress!')`.

Waffle isn't making the connection, but this fixes #42.

### DIFF
--- a/chrome/content/binding.xml
+++ b/chrome/content/binding.xml
@@ -13,4 +13,14 @@
       </xul:hbox>
     </content>
   </binding>
+
+  <binding id="recommendation-urlbar" extends="chrome://browser/content/urlbarBindings.xml#urlbar">
+    <handlers>
+      <handler event="keypress"><![CDATA[
+        // for now, just call the handler method defined on the parent binding.
+        // we'll add our own key handling logic in #43, #44.
+        return this.handleKeyPress(event);
+      ]]></handler>
+    </handlers>
+  </binding>
 </bindings>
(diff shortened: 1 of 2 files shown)


chuckharmston:
(Waffle uses commit messages to associate; amend it to add "(closes #42)" and it should get there.)

chuckharmston:
:+1: 

A question for me: to what is `this` in the `<handler>` bound?

Issue #54, opened 2016-03-01
[object Object] in recommendation abort message


jaredhirsch:
Now that we've got printable keys going, we can see the recommendation server abort in-flight xhrs when another keystroke arrives, and it looks like the error message could be improved:

<img width="790" alt="screen shot 2016-03-01 at 2 53 36 pm" src="https://cloud.githubusercontent.com/assets/96396/13444818/693f2d78-dfbd-11e5-8345-16937637d662.png">

(Ignore the `this.mInput is null` error; that's already captured in #29.)

The task here is to figure out why the error message looks the way it does, and then figure out how to make it more meaningful.

jaredhirsch:
This was fixed by #55, but I forgot the additional "fixes" in the commit message. Closing.

PR #55, opened 2016-03-01
Detect printable keys in urlbar and fire event

jaredhirsch:
@chuckharmston R?

This PR should be pretty straightforward, aside from using the Gecko-specific `KeyboardEvent.key` to decide if a key is printable or not.

You can test that this is working by pasting this into the Browser Toolbox console:

``` js
window.universalSearch.events.subscribe('printable-key', (e) => { console.log(JSON.stringify(e)) })
```

Then, as you type in the urlbar, you'll see the event fired if you typed a printable character.

I'm not sure if the event name should be `search-query` or `urlbar-updated`, or something else? One thing's for sure, the contents of the signal don't include a single printable key, so the term `printable-key` shouldn't be used. Suggestions welcome. APIs are hard :-P

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,3 +43,21 @@ By participating in this project, you agree to abide by our [code of conduct](./
 - event naming convention: hyphenated lower case.
 	- we don't need to namespace it to individual items because it's such a small list of events. just keep it simple.
 
+## Pubsub events
+
+Events that flow through the global event broker (`app.events`, invoked as
+`this.events` by individual modules):
+
+#### `urlbar-change`
+
+When the user types a printable key in the urlbar, the complete string in
+the urlbar is fired as the body of this event.
+
+```
+{
+  query: 'the mar'
+}
+```
+
+- query: unicode string contents of the urlbar
+
--- a/chrome/content/binding.xml
+++ b/chrome/content/binding.xml
@@ -17,9 +17,9 @@
   <binding id="recommendation-urlbar" extends="chrome://browser/content/urlbarBindings.xml#urlbar">
     <handlers>
       <handler event="keypress"><![CDATA[
-        // for now, just call the handler method defined on the parent binding.
-        // we'll add our own key handling logic in #43, #44.
-        return this.handleKeyPress(event);
+        // If the urlbar handles the event, it returns true. Else, it returns
+        // false, and the existing XBL key handlers handle the key event.
+        return window.universalSearch.urlbar.onKeyPress(event) || this.handleKeyPress(event);
       ]]></handler>
     </handlers>
   </binding>
--- a/lib/recommendation-server.js
+++ b/lib/recommendation-server.js
@@ -70,7 +70,7 @@ RecommendationServer.prototype = {
       this.events.publish('recommendation', data);
     }, reason => {
       // TODO: send this information to metrics instead of logging
-      console.error(reason);
+      this.console.error(reason);
     });
   },
   _search: function(query) {
--- a/lib/recommendation-server.js
+++ b/lib/recommendation-server.js
@@ -38,13 +38,13 @@ RecommendationServer.prototype = {
     this.request.isInFlight = false;
     this.warm();
 
-    this.events.subscribe('printable-key', this.search);
+    this.events.subscribe('urlbar-change', this.search);
   },
   destroy: function() {
     this.request.abort();
     delete this.request;
 
-    this.events.unsubscribe('printable-key', this.search);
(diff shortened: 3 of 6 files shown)


jaredhirsch:
Hmm, I think we should really just do `return urlbar.onKeyPress(event) || this.handleKeyPress(event)`, and not spread logic between JS and XBL. Going to update the patch now.

README.md
```diff
@@ -43,3 +43,21 @@ By participating in this project, you agree to abide by our [code of conduct](./
...
+
+#### `printable-key`
+
+When the user types a printable key in the urlbar, the complete string in
+the urlbar is fired as the body of the event.
+
+```
+{
+  query: 'the mar'
+}
```
chuckharmston, line 59: Should we also send over the full text of what the urlbar attempts to complete with? e.g. "facebook.com" here when pressing "f":

screen shot 2016-03-02 at 10 36 18 am

chrome/content/binding.xml
```diff
@@ -17,9 +17,9 @@
   <binding id="recommendation-urlbar" extends="chrome://browser/content/urlbarBindings.xml#urlbar">
     <handlers>
       <handler event="keypress"><![CDATA[
-        // for now, just call the handler method defined on the parent binding.
-        // we'll add our own key handling logic in #43, #44.
-        return this.handleKeyPress(event);
+        // If the urlbar handles the event, it returns true. Else, it returns
+        // false, and the existing XBL key handlers handle the key event.
+        return window.universalSearch.urlbar.onKeyPress(event) || this.handleKeyPress(event);
```
chuckharmston, line 22: :+1:

lib/ui/urlbar.js
```diff
@@ -7,25 +7,60 @@ const EXPORTED_SYMBOLS = ['Urlbar'];
...
+    // Criteria applied to decide if a key is printable:
+    //
+    // 1. Based on mdn.io/KeyboardEvent/key#Key_values, any printable key
+    //    will have an event.key value of length 1, that is, just the char
+    //    itself. Other keys will have longer string values, like 'Escape'.
+    //
+    // 2. If modifiers other than Shift are pressed, the user might be typing
+    //    a shortcut key combination. So, if those are pressed, do nothing.
+    //
+    // 3. If we're in private browsing mode, respect that and do nothing.
```
chuckharmston: Not sure this is the correct place for this; seems like a separate concern. Maybe as an `if` block wrapping `handlePrintableKey`?

chuckharmston:
> I'm not sure if the event name should be `search-query` or `urlbar-updated,` or something else? One thing's for sure, the contents of the signal don't include a single printable key, so the term `printable-key` shouldn't be used. Suggestions welcome. APIs are hard :-P

Let's mimic `<input>` elements and use `urlbar-change`.

chuckharmston:
Looks great, merge as you deem appropriate.

chuckharmston:
#54 should be addressed here; it's a result of the event binding you created in this PR. Update `RecommendationServer.prototype.search` to look like this:

``` javascript
search: function(payload) {
  this._search(payload['query']).then(data => {
```

And it should be good to go. You may want to update the commit message to close #54 as well.

README.md
jaredhirsch, line 59: We definitely could, what did you have in mind?

chuckharmston, line 59: It seems like something we'd want to log for metrics, so we could determine e.g. how many times f => facebook.

lib/ui/urlbar.js
6a68: Yup, have a look at the `onKeyPress` method above, that's just what the code does now; I moved the logic over from XBL

chuckharmston: Yeah, it definitely works; my point was that private browsing mode doesn't technically affect whether a key is printable or not.

But that's an academic concern, don't worry about it much.

jaredhirsch: Derp, did you mean the privacy part? Like, make it a first-class separate check? That sounds good.

README.md
jaredhirsch, line 59: For reference, filed as #56

lib/recommendation-server.js
```diff
@@ -64,12 +64,12 @@ RecommendationServer.prototype = {
       req.send();
     }
   },
-  search: function(query) {
-    this._search(query).then(data => {
+  search: function(payload) {
+    this._search(payload.query).then(data => {
```
jaredhirsch, line 69: THanks for pointing out the #54 fix, @chuckharmston :-)

Issue #58, opened 2016-03-02
Add a preferences observer to recommendation-server.js


jaredhirsch:
The recommended way to track preferences updates is to use a preferences observer. Currently RecommendationServer._url rechecks the value of the `pref each time it's called (potentially on every keystroke).

This is a performance enhancement, but one to consider sooner than later.

PR #60, opened 2016-03-02
Remove default preferences and any assumptions that preferences are set (temporary fix for #58).

chuckharmston:
The correct approach will be to add an observer, but this will help us through some debugging for now.


jaredhirsch:
Yup yup!

PR #61, opened 2016-03-02
Injects recommendation server result into popup (closes #57).

### LINKED ISSUE #57: Add recommendation to hbox when received from server.

chuckharmston:
screen shot 2016-03-02 at 3 47 09 pm

Not ready to merge yet; it needs some cleanup and docs, but the core/approach should be stable and could use a review overnight.

r? @6a68

### DIFF
--- a/lib/ui/recommendation.js
+++ b/lib/ui/recommendation.js
@@ -3,11 +3,78 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const EXPORTED_SYMBOLS = ['Recommendation'];
+const ELEMENTS = {
+  'RECOMMENDATION': 'universal-search-recommendation',
+  'CONTAINER': 'universal-search-recommendation-container',
+  'IMAGE': 'universal-search-recommendation-container',
+  'TITLE': 'universal-search-recommendation-title',
+  'URL': 'universal-search-recommendation-url',
+  'LABEL': 'universal-search-recommendation-label'
+};
+
+
+function RecommendationRow(win, data) {
+  this.win = win;
+  this.doc = win.document;
+  this.data = data;
+}
+
+
+RecommendationRow.prototype = {
+  render: function() {
+    let row = this;
+    let container = this.doc.createElement('box');
+    container.setAttribute('id', ELEMENTS.CONTAINER);
+    [
+      this.renderImage, this.renderTitle,
+      this.renderUrl, this.renderLabel
+    ].forEach(fn => {
+      const content = fn(row);
+      if (content) {
+        container.appendChild(content);
+      }
+    });
+    return container;
+  },
+
+  renderImage: function(row) {
+    if (row.data.enhancements.hasOwnProperty('domain')) {
+      let image = row.doc.createElement('image');
+      image.setAttribute('id', ELEMENTS.IMAGE);
+      image.setAttribute('src', row.data.enhancements.domain.logo);
+      return image;
+    }
+    return null;
+  },
+
+  renderTitle: function(row) {
+    let title = row.doc.createElement('hbox');
+    title.setAttribute('id', ELEMENTS.TITLE);
+    title.textContent = row.data.result.title;
+    return title;
+  },
+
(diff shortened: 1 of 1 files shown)


lib/ui/recommendation.js
```diff
@@ -4,6 +4,60 @@
...
+  this.data = data;
+}
+
+
+RecommendationRow.prototype = {
+  render: function() {
+    let row = this.win.document.createElement('box');
+
+    let image = this.renderImage();
+    if (!!image) {
```
jaredhirsch: elements are truthy, so you don't need the double-bang

lib/ui/recommendation.js
```diff
@@ -4,6 +4,60 @@
...
+    if (this.data.enhancements.hasOwnProperty('domain')) {
+      let image = this.win.document.createElement('image');
+      image.setAttribute('src', this.data.enhancements.domain.logo);
+      return image;
+    }
+    return null;
+  },
+
+  rendertitle: function() {
+    let title = this.win.document.createElement('hbox');
```
jaredhirsch: I'm guessing there are some other XUL element types that we could use, rather than hbox everywhere. This is fine for now, though; maybe file a followup bug for after the 3/15 milestone?

lib/ui/recommendation.js
```diff
@@ -4,6 +4,60 @@
...
+  renderImage: function() {
+    if (this.data.enhancements.hasOwnProperty('domain')) {
+      let image = this.win.document.createElement('image');
+      image.setAttribute('src', this.data.enhancements.domain.logo);
+      return image;
+    }
+    return null;
+  },
+
+  rendertitle: function() {
```
jaredhirsch: `renderTitle` for consistency?

lib/ui/recommendation.js
```diff
@@ -48,16 +104,25 @@ Recommendation.prototype = {
...
+    this.el.collapsed = false;
+    this.events.publish('recommendation-shown', this);
   },
+
   hide: function() {
-    // collapse the node
-    // null out this.data
+    this.win.universalSearch.console.log('recommendation.hide');
+    this.el.collapsed = true;
+    while (this.el.firstChild) {
```
jaredhirsch: It should be enough to just pop the row element off of `this.el`, as the other nodes are all child nodes of the row. This works too, but it should only ever run once.

jaredhirsch:
@chuckharmston This looks great, nothing like being stuck in an XML document to make `innerHTML` seem like the best API ever :-P

Picked at a few things above, but really, this looks great. Happy to help with another look tomorrow, or feel free to land it if it seems to work. :beers:

chuckharmston:
> This looks great, nothing like being stuck in an XML document to make innerHTML seem like the best API ever :-P

I actually couldn't get innerHTML to work; it seemed to have issues with both namespaced and un-namespaced XML. There's probably a fix, but this isn't particularly terrible itself.

jaredhirsch:
> I actually couldn't get innerHTML to work; it seemed to have issues with both namespaced and un-namespaced XML. There's probably a fix, but this isn't particularly terrible itself.

I looked at MDN for about 20 minutes...there doesn't seem to be an `innerHTML` attribute available on any of the XUL elements, I'm guessing because of the different doctype :sob:

jaredhirsch:
BTW here's the most amusing of the oddities I found in the chamber of deprecated horrors in MDN: look at the example "reverse JXON algorithm", specifying XML contents via some bizarro JSON schema!

https://developer.mozilla.org/en-US/docs/Web/API/Document_object_model/How_to_create_a_DOM_tree

lib/ui/recommendation.js
chuckharmston: Very much open to suggestion, but I didn't see anything remotely helpful in the [XUL reference](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/XUL_Reference).

chuckharmston: https://github.com/mozilla/universal-search/issues/63

PR #62, opened 2016-03-02
Intercept navigational keys if the recommendation is visible

jaredhirsch:
@chuckharmston R?

Interesting changes here:
- We have to set `phase="capturing"` on the XBL key handler to intercept the key before it is handled by the existing code. Gotta love those leftover decisions from the '90s browser wars...
- Had to adjust the urlbar/recommendation relationship to avoid circular constructor injection dependencies. If this gets any more annoying, I'm all for just passing an `app` reference around in the constructors. The current approach is: use events for one-way communication; use callbacks/promises for two-way communication. Thoughts/comments welcome.
- Updated the recommendation view docs to better explain why we're polling the XUL DOM.
- I'm not happy with the `navigational-key` event name, although it's reasonably descriptive. Suggestions very welcome.

### DIFF
--- a/chrome/content/binding.xml
+++ b/chrome/content/binding.xml
@@ -16,7 +16,7 @@
 
   <binding id="recommendation-urlbar" extends="chrome://browser/content/urlbarBindings.xml#urlbar">
     <handlers>
-      <handler event="keypress"><![CDATA[
+      <handler event="keypress" phase="capturing"><![CDATA[
         // If the urlbar handles the event, it returns true. Else, it returns
         // false, and the existing XBL key handlers handle the key event.
         return window.universalSearch.urlbar.onKeyPress(event) || this.handleKeyPress(event);
--- a/lib/ui/recommendation.js
+++ b/lib/ui/recommendation.js
@@ -73,7 +73,6 @@ RecommendationRow.prototype = {
 function Recommendation(opts) {
   this.win = opts.win;
   this.events = opts.events;
-  this.urlbar = opts.urlbar;
   this.el = null;
 
   this.navigate = this.navigate.bind(this);
@@ -84,10 +83,10 @@ function Recommendation(opts) {
 
 Recommendation.prototype = {
   init: function() {
-    // Note: there is a race between XBL modifying the XUL DOM to insert our
-    // this.el, and the scripts being loaded and initialized. Rather than try
-    // to fire a 'ready' event / set app state from the XBL <constructor>,
-    // which is sometimes unreliable, we instead poll the XUL DOM for the el.
+    // Note: the XUL DOM isn't updated to insert the
+    // 'universal-search-recommendation' element until the popup is opened
+    // once. So, poll for the element until it's ready, then attach listeners
+    // to it.
     this._pollForElement();
 
     this.events.subscribe('recommendation', this.show);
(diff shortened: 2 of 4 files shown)


chrome/content/binding.xml
```diff
@@ -16,7 +16,7 @@
 
   <binding id="recommendation-urlbar" extends="chrome://browser/content/urlbarBindings.xml#urlbar">
     <handlers>
-      <handler event="keypress"><![CDATA[
+      <handler event="keypress" phase="capturing"><![CDATA[
```
chuckharmston, line 19: burn it down

lib/ui/recommendation.js
```diff
@@ -17,10 +16,10 @@ function Recommendation(opts) {
...
 Recommendation.prototype = {
   init: function() {
-    // Note: there is a race between XBL modifying the XUL DOM to insert our
-    // this.el, and the scripts being loaded and initialized. Rather than try
-    // to fire a 'ready' event / set app state from the XBL <constructor>,
-    // which is sometimes unreliable, we instead poll the XUL DOM for the el.
+    // Note: the XUL DOM isn't updated to insert the
+    // 'universal-search-recommendation' element until the popup is opened
+    // once. So, poll for the element until it's ready, then attach listeners
+    // to it.
```
chuckharmston, line 89: You know, polling is actually a lot less hacky when the element isn't injected until the pop is opened.

chuckharmston, line 89: We could probably up the polling time to 50 or 75 ms, since it won't prevent queries and the round trip time will likely take at least that long.

lib/ui/recommendation.js
```diff
@@ -61,7 +59,7 @@ Recommendation.prototype = {
   navigate: function(evt) {
     // if it's a click, we'll have a MouseEvent; if it's a right-click, bail.
     // else, we'll just have the (possibly null) data from the 'enter-key' event.
-    // call this.urlbar.navigate to trigger navigation, passing in the URL.
-    // TODO: or should we fire an event and keep all the modules maximally disconnected?
+    // fire a 'navigate-requested' event (please find a better name) to trigger navigation.
```
chuckharmston: Thinking about event names. Let's use the form `prefix-verb` with three different prefixes:
- `urlbar` for any events in the urlbar itself.
- `recommendation` for any events related to our hbox
- `awesomebar` for any events related to the awesomebar that are not our hbox 

So here, maybe `recommendation-navigate`?

lib/ui/urlbar.js
```diff
@@ -64,6 +69,17 @@ Urlbar.prototype = {
       this.events.publish('urlbar-change', data);
     });
   },
+  _navigationalKeys: ['ArrowUp', 'ArrowDown', 'PageUp', 'PageDown', 'Tab'],
```
chuckharmston, line 78: At least we don't have to work with key codes.

chuckharmston:
Looks great! Merge at will.

lib/ui/recommendation.js
jaredhirsch, line 89: :+1:

lib/ui/recommendation.js
jaredhirsch: sure, sounds good. I'll add this to the README docs, too. thanks!

PR #65, opened 2016-03-03
Retries injection of first recommendation if recommendation row not yet in popup (closes #64).

### LINKED ISSUE #64: First recommendation is not shown
There appears to be a race condition between the discovery of the `#universal-search-recommendation` element and `Recommendation.show` being run for the first recommendation, which prevents it from being inserted to the DOM.

chuckharmston:
This really sucks, but it works. `¯\_(ツ)_/¯`

r? @6a68

### DIFF
--- a/lib/ui/recommendation.js
+++ b/lib/ui/recommendation.js
@@ -119,11 +119,17 @@ Recommendation.prototype = {
   },
 
   show: function(data) {
-    const recommendation = new RecommendationRow(this.win, data);
-    this.el.appendChild(recommendation.render());
-    this.el.collapsed = false;
-    this.data = data;
-    this.events.publish('recommendation-shown', this);
+    if (this.el) {
+      const recommendation = new RecommendationRow(this.win, data);
+      this.el.appendChild(recommendation.render());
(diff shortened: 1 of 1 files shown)


jaredhirsch:
@chuckharmston Hmm. In cases where there are dueling timeouts (polling for this.el, and separately waiting for this.el to appear, to insert content in it), it can be nicer to store a promise on the object that resolves when one of the timeouts is done, so we can ensure events run in the order we want.

However, in this case, I kinda think we should err on the side of showing nothing if the element isn't available yet. If we wait a turn, the JS thread yields to the UI thread, and we have no idea what the state of the world will be when JS resumes: the search query might be different, the popup might have been closed. We could track that state by assigning the timeout ID on the object and clearing the timeout when a new query is issued--but now we're talking about a lot of extra state. Maybe we just don't sweat the first recommendation for the first keystroke.

In fact, now that I think about it, the existing search code doesn't even ping for search suggestions until the second key is typed. (Maybe this is part of the reason why!)

chuckharmston:
Fair enough! Closing.

PR #66, opened 2016-03-03
Styles inserted recommendation (closes #5).

### LINKED ISSUE #5: style our hbox element
notes:
- need a highlight style that matches the richlistbox
- we'll use `hbox.collapsed = true` to hide elements in JS, no need to create a hidden class

chuckharmston:
screen shot 2016-03-03 at 11 12 36 am

A video of it in action: https://chchc.ch/the-mart.mov

r? @6a68

### DIFF
--- a/lib/ui/recommendation.js
+++ b/lib/ui/recommendation.js
@@ -107,6 +107,7 @@ function Recommendation(opts) {
   this.events = opts.events;
   this.urlbar = opts.urlbar;
   this.el = null;
+  this.timeout = null;
 
   this.navigate = this.navigate.bind(this);
   this.show = this.show.bind(this);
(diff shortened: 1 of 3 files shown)


lib/ui/recommendation.js
```diff
@@ -44,12 +57,26 @@ RecommendationRow.prototype = {
...
       return image;
     }
-    return null;
+    return row.renderLetterbox(row);
+  },
+
+  _firstLetterInDomain: function(url) {
+    const domain = url.replace(/^[^\/]+\/+/, '').split('/')[0].split('.');
+    const firstLevelDomain = domain[domain.length - 2];
+    return firstLevelDomain[0];
```
jaredhirsch: Simpler, more reliable way to get the initial letter of the domain name:

``` js
// first, get an nsIURI from the url
let uri = Services.ios.newURI(url, null, null);
// next, use the TLD service to get the base domain
let baseDomain = Services.eTLD.getBaseDomain(uri);
let initialLetter = baseDomain[0];
```

The advantage of using the existing services is that they account for all the quirks / edge cases in global domain names and TLDs.

jaredhirsch:
@chuckharmston This looks fantastic, great work. One suggestion with the domain letter thing above, other than that, feel free to merge when you're ready.

clouserw:
sweet! :100:

chuckharmston:
Updated screencap with massive jank reduction: https://chchc.ch/with-timeout.mov

This is achieved by setting a 50ms timeout on hiding a result when it's obsolete; then, if that timeout hasn't executed yet, we cancel it and replace the old recommendation with the new one. 50ms may not be the correct number when executing across a network, but it's good enough for now and is easily tweakable.

PR #73, opened 2016-03-08
WIP key focus handling (fixes #38)

### LINKED ISSUE #38: highlight manager: adjust highlight on 'navigational-key' event
#43 covers the part where the urlbar intercepts the keypress, and, if the recommendation is visible, fires a `navigational-key` event.

This bug covers the highlight management from that point onward: given a recommendation and a list of results, move the highlight to the right item.

The two pieces of state we care about are whether the recommendation is highlighted (TBD property name), and whether the richlistbox is (available via `gURLBar.popup.selectedIndex`, if the selectedIndex is -1, then nothing's highlighted).

The down, page down, or tab keys move downward in the popup, and the up, page up, and shift+tab keys move upward.

One edge case to consider: keying down or up past the end of the list causes the highlight to 'wrap around'.

jaredhirsch:
@chuckharmston R? Have a little more cleanup to do, and the comments aren't all quite correct, but the focus management works reliably, with very occasional blue flickering in the top result row.

Most of the complexity here is due to the fact that the underlying autocomplete.xml code doesn't insert 30 results at once; instead, it sets 5 timeouts that insert 6 rows each. (The intention, according to a comment in the code, is to keep the urlbar responsive by batching updates.) The problem is, if we set a timeout to steal the focus, our timeout winds up behind the others on the stack, each of which causes tons of DOM updates, so that there's a long lag between the blue highlight being applied to the top result item, and our code unhighlighting that item. 

To hack around this, we (1) use a MutationObserver to detect when the rows are changing or being inserted, and (2) in response to those mutation events, we steal focus three times: synchronously, on the next frame using `requestAnimationFrame`, and two frames ahead, via a nested `requestAnimationFrame` call.

(This branch seems amazingly performant if you revert the dummy data commit. I'm kinda tempted to revert that in master, might open a PR to discuss tomorrow.)

There are some smaller changes, some of which need comments, but don't yet have them. I'll add comments before this lands; please call out anything that seems unclear. The squashed commit messages below describe the smaller changes:
- Don't try to show/hide if the recommendation el doesn't yet exist
- Separately handle grabbing the highlight when the rec is shown
- Correct object graph initialization order
- Track both the before and after popup hiding events
- We may sometimes need to prevent the popup from hiding, by canceling
  the 'popuphiding' XUL event (our 'before-popup-hide' event). In such
  cases, we don't want to hide the recommendation unless the popup is
  really hidden. XUL exposes the 'popuphidden' event, which we can hook
  into as an 'after-popup-hide' event.
- Remove mouse handlers from highlight manager for now, too much complexity
- Avoid referring to gURLBar; we already have 'this.popup'
- Always handle nav keys, now that we are replacing the XBL focus logic completely
- send the recommendation-shown event after a turn, so it's in the dom
- use mutation observer to steal focus over and over as rows are
  inserted into the popup
- use 2 requestAnimationFrames to minimize the amount of blue highlight
  flickering on the zeroth result list item as results stream in.
- take an initial swing at the highlight style

### DIFF
--- a/chrome/content/binding.xml
+++ b/chrome/content/binding.xml
@@ -4,6 +4,13 @@
           xmlns:html="http://www.w3.org/1999/xhtml"
           xmlns:xul="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
           xmlns:xbl="http://www.mozilla.org/xbl">
+
+  <![CDATA[
+    /*
+     * Override the default binding, inserting a recommendation element in
+     * the popup above the list of history/suggestion results.
+     */
+  ]]>
   <binding id="recommendation-popup" extends="chrome://browser/content/urlbarBindings.xml#urlbar-rich-result-popup">
     <content ignorekeys="true" level="top" consumeoutsideclicks="never" aria-owns="richlistbox">
       <xul:hbox id="universal-search-recommendation" collapsed="true"/>
@@ -12,12 +19,98 @@
         <children/>
       </xul:hbox>
     </content>
+    <implementation>
+      <![CDATA[
+        /*
+         * The selectedIndex setter sets the blue highlight on the currently-
+         * selected item in the results list. The existing implementation,
+         * defined in autocomplete.xml, automatically closes the popup if no
+         * results list items are selected (selectedIndex = -1).
+         *
+         * Override the setter, removing the auto-close behavior, so that the
+         * universal search recommendation element can steal the highlight.
+         */
+      ]]>
+      <field name="_selectedIndex">-1</field>
+      <property name="selectedIndex">
+        <getter>
+          return this._selectedIndex;
+        </getter>
+        <setter><![CDATA[
+          var resultsContainer = document.getAnonymousElementByAttribute(gURLBar.popup, 'anonid', 'richlistbox');
+
+          // NOTE: getElementsByClassName is needed to obtain a live NodeList.
+          // The results aren't inserted all at once: they are inserted a few
+          // at a time over many successive turns. In order to avoid setting
+          // the selectedIndex past the end of the results list, and also to
+          // avoid mistakenly concluding that the results list is empty, the
+          // NodeList needs to be live, not static.
+          var resultRows = resultsContainer.getElementsByClassName('autocomplete-richlistitem');
+
+          if (!resultRows || !resultRows.length) {
+            return;
+          }
+
+          if (val > resultRows.length - 1) {
+            val = resultRows.length - 1;
+          } else if (val < -1) {
+            val = -1;
+          }
+
+          // Unselect all the rows (in case of any weird multiple-focus bugs),
+          // then immediately select the appropriate row, to hopefully batch
+          // the selection changes into a single DOM update.
+          Array.prototype.forEach.call(resultRows, function(row, i) {
+            row.selected = false;
+          });
+
+          // Highlight the row, if it's found, else highlight nothing.
+          if (resultRows[val]) {
+            resultRows[val].selected = true;
+            // Scroll the row into view, if necessary.
+            resultsContainer.ensureIndexIsVisible(val);
+            this._selectedIndex = val;
+            return val;
+          } else {
+            this._selectedIndex = -1;
+            return -1;
+          }
+        ]]></setter>
+      </property>
+
+      <![CDATA[
+        /*
+         * The urlbar stops working if getNextIndex is not overridden here.
+         * This implementation doesn't handle certain edge cases, like
+         * scrolling off the top or bottom of the list, but whatever code
+         * calls this code doesn't seem to hit that bug.
+         */
+      ]]>
+      <method name="getNextIndex">
+        <parameter name="reverse"/>
+        <parameter name="amount"/>
+        <parameter name="index"/>
+        <parameter name="maxRow"/>
+        <body><![CDATA[
+          let realAmount = amount % maxRow;
+          this._selectedIndex = reverse ? this._selectedIndex - realAmount: this._selectedIndex + realAmount;
+          return this._selectedIndex;
+        ]]></body>
+      </method>
+    </implementation>
   </binding>
 
+  <![CDATA[
+    /*
+     * Override the urlbar so that the universal search code can conditionally
+     * intercept key events before they are sent to the existing urlbar key
+     * handler code, defined in urlbarBindings.xml.
+     */
+  ]]>
   <binding id="recommendation-urlbar" extends="chrome://browser/content/urlbarBindings.xml#urlbar">
     <handlers>
       <handler event="keypress" phase="capturing"><![CDATA[
-        // If the urlbar handles the event, it returns true. Else, it returns
+        // If our add-on handles the event, it returns true. Else, it returns
         // false, and the existing XBL key handlers handle the key event.
         return window.universalSearch.urlbar.onKeyPress(event) || this.handleKeyPress(event);
       ]]></handler>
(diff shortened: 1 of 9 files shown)


lib/ui/recommendation.js
```diff
@@ -162,10 +181,28 @@ Recommendation.prototype = {
...
     this.el.collapsed = false;
     this.data = data;
-    this.events.publish('recommendation-shown', this);
+    // We actually need to wait a turn for highlight adjustment to succeed,
+    // assuming a flow of events like this:
+    // 1. ask the DOM to render the element
+    // 2. wait a turn for the UI thread to render it into the DOM
+    // 3. adjust highlight once everything's in the DOM
+    //
+    // Because the XBL code calls setTimeout 30 times for each set of results,
```
jaredhirsch: this comment is wrong, setTimeout is called up to 5 times

chrome/content/binding.xml
```diff
@@ -12,6 +12,79 @@
         <children/>
       </xul:hbox>
     </content>
+    <implementation>
+      <field name="_selectedIndex">-1</field>
```
jaredhirsch, line 34: All this selectedIndex code is needed because we needed to override the existing selectedIndex setter. It's such a convoluted story that I've forgotten exactly why, but I'll piece it together and insert a comment before this lands.

chuckharmston:
screen shot 2016-03-08 at 11 23 57 am

By typing fast I'm sometimes able to get a double-selection to happen.

chuckharmston:
Of course now that I'm trying to reproduce it I can't, but the first two times I ran your branch I was able to get Firefox to lock up (spinning beachball-style). Nothing to do here, just worth monitoring.

lib/ui/highlight-manager.js
```diff
@@ -9,36 +9,143 @@ function HighlightManager(opts) {
...
+    //
+    // The existing XBL code wraps each insertion in a setTimeout, so we can't
+    // rely on setTimeout to steal the highlight: if 30 row rendering calls are
+    // in the event loop ahead of this code, the blue highlight will very
+    // noticeably flicker on the top result row. To avoid this, we steal the
+    // highlight immediately, then use rAF to steal it just before the next
+    // paint, then set *another* rAF inside the first rAF, preventing another
+    // paint from showing a result in the list as highlighted. This seems a
+    // little excessive, but actually the inner rAF seems to be the key to
+    // truly minimizing the flickering as we steal the highlight.
```
chuckharmston: I have no idea why this works, but :tada: :balloon: :dizzy:

chuckharmston: I have no idea why this works, but :tada::balloon::dizzy:

lib/ui/highlight-manager.js
```diff
@@ -9,36 +9,143 @@ function HighlightManager(opts) {
...
+    const selectedIndex = this.popup.selectedIndex;
+    const recommendationVisible = this.recommendation.el && !this.recommendation.el.collapsed;
+    const recommendationHighlighted = this.recommendation.el && this.recommendation.isHighlighted;
+
+    // Clear all highlights. The DOM is now dirty and not trustworthy.
+    this.clearHighlight();
+
+    // Start by checking what's currently highlighted, and alter the state:
+
+    // recommendation
```
chuckharmston: Nit: can we refactor comments to start with a capital letter and be a complete sentence?

lib/ui/recommendation.js
```diff
@@ -162,10 +181,28 @@ Recommendation.prototype = {
...
+    // the timeouts sometimes seem to take a long time to fire. Let's avoid
+    // waiting on the overloaded event loop by using rAF, and firing the event
+    // just before the next browser paint.
+    //
+    // TODO: should we just fire this event immediately, and let the highlight
+    // manager handle all of the async UI twiddling?
```
chuckharmston: This TODO makes a lot of sense to me.

chrome/skin/style.css
```diff
@@ -85,3 +85,13 @@ textbox#urlbar {
 #universal-search-recommendation-label {
   color: #BBB;
 }
+
+/* highlight color, applied when the recommendation is highlighted */
+
+#universal-search-recommendation.highlight {
+  background-color: #3b8ce3;
```
chuckharmston: Can we use the inbuilt CSS here instead for a little more native-feeling UI? It looks like the appropriate highlight styles are applied via `.autocomplete-richlistitem[selected="true"]`:

https://dxr.mozilla.org/mozilla-central/source/toolkit/themes/linux/global/autocomplete.css#110
https://dxr.mozilla.org/mozilla-central/source/toolkit/themes/osx/global/autocomplete.css#96
https://dxr.mozilla.org/mozilla-central/source/toolkit/themes/windows/global/autocomplete.css#104
https://dxr.mozilla.org/mozilla-central/source/toolkit/themes/windows/global/autocomplete.css#112

lib/ui/highlight-manager.js
```diff
@@ -9,36 +9,143 @@ function HighlightManager(opts) {
...
     delete this.popup;
     delete this.recommendationView;
     delete this.win;
   },
+  mutationHandler: function() {
+    if (this.recommendation.el && !this.recommendation.el.collapsed) {
+      this.stealHighlight();
+    }
+  },
+  stealHighlight: function() {
```
chuckharmston, line 74: I really like this API. Clear, clean, obvious. Well done.

lib/ui/urlbar.js
```diff
@@ -85,9 +82,11 @@ Urlbar.prototype = {
     return this._navigationalKeys.indexOf(evt.key) > -1;
   },
   handleNavigationalKey: function(evt) {
+    // preventDefault on the event, otherwise it seems up and down move two
+    // steps at a time, some key handler way down in the XUL stack
```
chuckharmston, line 91: Hah, I bet that was slightly maddening to debug.

chuckharmston:
As we discussed, can you really aggressively document this? The XBL magic :fireworks: especially, but even adding docstrings explaining what each method is doing and why would be super helpful to future engineers.

chuckharmston:
This is madness, and I love it. If you can fix the double-highlight, use the CSS, and document it thoroughly, feel free to merge at will.

chrome/skin/style.css
jaredhirsch: that's a great idea

lib/ui/highlight-manager.js
jaredhirsch: It is the height of madness, that's for sure. I guess the time when the setTimeouts will be executed is unpredictable, so if we spray the event loop with enough focus stealing DOM changes, we wind up doing pretty well on average?

jaredhirsch:
Sounds great, thanks very much for taking a look :-)

chuckharmston:
It would be extra great if you could do two things:
1. If the recommendation is selected, set `bottom-border-color: transparent;`
2. If the item below the recommendation is selected, `position: relative; top: -1px;` and increase `padding-top` by one pixel to effectively hide the border-bottom on the recommendation.

chuckharmston:
Could you add the bug number/s in the PR title so Waffle can pick up on it?

jaredhirsch:
@chuckharmston To your point about the CSS:

> It would be extra great if you could do two things:
> 1. If the recommendation is selected, set `bottom-border-color: transparent;`
> 2. If the item below the recommendation is selected, `position: relative; top: -1px;` and increase `padding-top` by one pixel to effectively hide the border-bottom on the recommendation.

It seems like nothing breaks if there's no bottom-border on the recommendation, so I'm just going to set `border-bottom: none` to avoid the horizontal line.

chuckharmston:
Hmmm, I was thinking that it would change the height calculations. If not, then great!

jaredhirsch:
Closing in favor of a non-WIP branch

Issue #35, opened 2016-02-19
Build docker containers with circleci


clouserw:
Cloud Services' Dockerflow uses circleci to build docker containers (https://github.com/mozilla-services/Dockerflow).  Testpilot just committed a circle.yml that works, we might be able to steal most of that:

https://github.com/mozilla/testpilot/commit/87b7c6eb9915f69976b5bc0701509c47409bc66a
https://github.com/mozilla/testpilot/commit/86774eb9cfc6266e619e9fc9056abbcce28b2074

or there is an example in the Dockerflow link and @mostlygeek can help with questions

clouserw:
Sorry, wrong repository.  closing in favor of https://github.com/mozilla/universal-search-recommendation/issues/40

Issue #18, opened 2016-02-10
client-side metrics


jaredhirsch:
use nsIObserver idea-town metrics signals to end serialized objects over

jaredhirsch:
Google doc with draft of metrics we'd gather: https://docs.google.com/document/d/1-FeQS1bMAT1S9MfPB6sm6czsRXdNf3OI3PTUw3JHU18/edit

jaredhirsch:
Setting the label to 'medium' means that we should aggressively timebox the initial pass at metrics, just reporting the absolute minimum. If there's time, we can report additional details.

clouserw:
Should have something for this soon.

See also https://github.com/mozilla/universal-search-recommendation/issues/35

chuckharmston:
I'll work on defining this, but implementation is blocked on mozilla/testpilot#234.

PR #79, opened 2016-03-11
Adds metrics plan for add-on and recommendation server.

chuckharmston:
Nicer-reading version here: https://github.com/chuckharmston/universal-search/blob/18-metrics-plan/docs/metrics.md

r? @6a68 @clouserw @nchapman

Connected to #18, connected to #35.


metrics.md
```diff
@@ -0,0 +1,85 @@
...
+- The HTTP status code of the response.
+- The enhancers applied to that query (i.e. `recommendation.enhancements.keys()`)
+
+We do not wish to record individual users browsing habits; we're interested in aggregations of user behavior. Thus, we will not record any queries that:
+
+- Are longer than 20 characters.
+- Appear to begin with a protocol (`^[^\s]+\:\S`).
+- Appear to be a hostname (`^[^\s]+\.\S`)
+
+Additionally, on a TBD schedule we will discard outliers:
```
chuckharmston: I'm not sure if/how this is possible in the Heka/Redshift flow.

metrics.md
```diff
@@ -0,0 +1,85 @@
...
+## Client metrics
+
+Metrics gathered from the Universal Search add-on will be reported to the Test Pilot add-on, which will augment that information and relay it to Firefox's telemetry system via `submitExternalPing`. This is not currently implemented on the Test Pilot side; those efforts are encompassed by [issue #234](https://github.com/mozilla/testpilot/issues/234).
+
+
+### Events
+
+Universal Search will record and report each of the following events:
+
+- User selects a recommendation.
```
ghost: You have these broken down by event, but you have a lot of duplication (eg. time from start to end).  Wouldn't "event" just be a type identifier?  So you'd record something like:

```
us.totaltime
us.querylength
us.eventtype
us.enhancers (if applicable)
us.datatype (if applicable)
us.levenstein (if applicable)
```

and then when you visualize you'd break it down into whatever events you wanted?  I know this is just the same thing you're saying here, but I'm trying to picture this from the POV of recording the data.  (that will help build the right structures to record the data)

metrics.md
```diff
@@ -0,0 +1,85 @@
...
+
+Once a day, Universal Search will report an aggregation of activity from the previous day. That data will include:
+
+- The number of searches in the previous day.
+- The number of those searches that ended in each of the following ways:
+	- By visiting a recommendation.
+	- By entering and visiting a URL directly into the URL bar.
+	- By entering and visiting a search query into the URL bar.
+	- By using a suggested search term.
+	- By visiting a bookmark.
```
ghost: Is this (and the history) meaning a bookmark that shows up in the URL bar, or would clicking a bookmark in the bookmark menu do the same thing (that is to say, clicking a link with no interaction with universal search at all)

metrics.md
ghost: Me either.  We'll have to ask.

metrics.md
```diff
@@ -0,0 +1,85 @@
+# Universal Search Metrics
```
ghost: This is a really good start, thanks Chuck.  Can you identify what the most important KPI is?  (probably something with engagement ratio?)

metrics.md
chuckharmston: Yep, that's likely how it'd be implemented. I was trying to sketch the telemetry plan agnostic of that, except where it would dictate the collection strategy (on the server).

metrics.md
chuckharmston: Any idea who the correct person is to ask about this?

metrics.md
chuckharmston: Naively, this would specifically mean by clicking on a bookmark in the awesome bar. This is somewhat indirectly implied by the use of "those searches", with "search" being defined at the top.

metrics.md
chuckharmston: @clouserw Is the official Mozilla Engagement Ratio formula finished? Last I saw it was a draft in a Google Doc somewhere.

ghost: nope, still draft as far as I know.  The discussion for Fx is happening in https://bugzilla.mozilla.org/show_bug.cgi?id=1240849

metrics.md
```diff
@@ -0,0 +1,85 @@
...
+	- Time from search start to search completion.
+	- Length of query.
+- Use selects an item from history.
+	- Time from search start to search completion.
+	- Length of query.
+
+
+### Daily ping
+
+Once a day, Universal Search will report an aggregation of activity from the previous day. That data will include:
```
jaredhirsch: What if the experiments just send metrics to test pilot as they are generated? That lets experiments avoid the work of implementing web storage solutions to hold metrics data, and lets the test pilot addon and server negotiate an update frequency.

metrics.md
jaredhirsch: I'd ask in #datapipeline

clouserw: I talked to Rob.  Sounds like it's fairly trivial to purge data from Redshift, and slightly more difficult to purge from S3, but we could put a retention policy on S3 since it's just a stop on the way to Redshift anyway.

tl;dr this should be possible.

metrics.md
nchapman: Yeah I agree that it could be a single event like selected/navigated/closed (whatever the right way to think about this) with these attributes attached to it.

metrics.md
```diff
@@ -0,0 +1,85 @@
...
+	- Time from search start to search completion.
+	- Length of query.
+	- Levenstein distance between query and completed search term.
+- User selects a bookmark.
+	- Time from search start to search completion.
+	- Length of query.
+- Use selects an item from history.
+	- Time from search start to search completion.
+	- Length of query.
+
```
nchapman: Probably worth recording when the user chooses nothing. This will be especially important when we attempt to answer questions in our recommendations such as with a Weather card.

metrics.md
nchapman: Seems like a great idea.

metrics.md
nchapman: Marshall would have some good input on this as well from the policy side.

metrics.md
nchapman: I agree this is a great start! I think we will end up reporting the engagement ratio for all Test Pilot tests "automatically". The metrics that I think are most important for this test specifically are the click through rates of the recommendations and retention. I'm working on a summary doc right now to capture these. We can integrate into this doc as soon as that's done.

metrics.md
jaredhirsch: We can break down 'user chooses nothing' in a few ways:
- did the user hit `Escape`, `LeftArrow`, or `RightArrow` to dismiss the popup?
- did the user switch tabs or windows?
- did the user switch focus away from the Firefox application altogether?

Also, I think it's worth recording whether a user's selection involves a key or mouse event.

jaredhirsch:
@chuckharmston Looks like a good start, I added a few comments inline. Thanks for driving this.

Data retention and access controls aren't discussed, but probably should be. I could see legal/privacy making those decisions for us, but it'd be good to mention them here, since users might think omitting those details implies that we don't have rules in place.

Feel free to reassign to me if you want to do a second round of feedback.

metrics.md
chuckharmston: This was written assuming that we didn't have access to a UID, which would have make any sort of proportional reporting (i.e. x% of the time, y outcome) impossible. Turns out, Test Pilot will be giving us that, so we should send them as they're generated.

metrics.md
chuckharmston: Agreed. As I said above, I'm avoiding jumping to implementation until we have reporting APIs to develop against.

metrics.md
chuckharmston: IRC chat with @6a68: it's probably enough to know that it closed, and not how they closed it.

chuckharmston:
Updated to address comments and add KPIs with some visualization (thanks, Nick!).

r? @6a68 

(Feel free to merge if you're happy where it's at; I think it's ready.)

docs/metrics.md
```diff
@@ -0,0 +1,114 @@
...
+	- Time from search start to search completion.
+	- Length of query.
+	- Levenstein distance between query and completed search term.
+- User selects a bookmark.
+	- Time from search start to search completion.
+	- Length of query.
+- Use selects an item from history.
+	- Time from search start to search completion.
+	- Length of query.
+- The user terminates a search.
```
clouserw, line 41: I'd like to see the actual JSON object (as part of this documentation) you'd be sending as the payload in submitExternalPing() to track all of these.

docs/metrics.md
```diff
@@ -0,0 +1,114 @@
...
+To gain a better understanding of how users make use of the URL bar, the recommendation server will maintain logs of queries passed to it. For each request, we will record:
+
+- The query string itself.
+- The HTTP status code of the response.
+- The enhancers applied to that query (i.e. `recommendation.enhancements.keys()`)
+
+We do not wish to record individual users browsing habits; we're interested in aggregations of user behavior. Thus, we will not record any queries that:
+
+- Are longer than 20 characters.
+- Appear to begin with a protocol (`^[^\s]+\:\S`).
```
clouserw, line 61: having the actual regex is great!

docs/metrics.md
```diff
@@ -0,0 +1,114 @@
...
+![](images/kpi_1.png)
+
+
+#### User retention
+
+_Do users continue to use recommendations? As we improve the recommendations, are they more likely to stick around?_
+
+This will allow us to understand overall retention. A user is considered retained if they use at least one recommendation each week.
+
+![](images/kpi_2.png)
```
clouserw, line 93: These images really help. Nice work on them

clouserw:
aside from the JSON question, I think it looks great. thanks chuck

docs/metrics.md
chuckharmston, line 41: Yep, Nick requested the same thing. In this PR, I'm just talking about the plan; the actual JSON blob will come with implementation when mozilla/testpilot#479 lands and I have actual APIs to develop against.

jaredhirsch:
@chuckharmston :tada: great stuff, let's document!

Issue #16, opened 2016-02-10
allow the popup to be pinned open


jaredhirsch:
Basically, we need a slightly fancier version of this one-liner:

``` js
gURLBar.popup.addEventListener('popuphiding', (e) => { e.preventDefault(); })
```

We'll need to:
- check some global app state to decide whether to pin the popup open
- not use an anonymous function, so the listener can be detached

chuckharmston:
Can we set this as addon config?

jaredhirsch:
it'd be a bit awkward, because the popup is pinned open per-window. enabling it in addon config would lead to a bunch of odd windows stuck open, and it might even be hard to resize the window enough to get access to the config screen to disable it ;-)

jaredhirsch:
hmm, maybe we could do a 'debug' config setting that displays a menu button that lets you pin/unpin the popup for a given window? that might be nice

chuckharmston:
We could always allow a pinned window to be closable through a button added in debug mode, a key sequence (`<esc> <esc>`), or something along those lines.

jaredhirsch:
:+1: all great ideas

However this gets implemented, we should also document the existence of the option in the README.

PR #80, opened 2016-03-11
Pass highlight between recommendation and results list

### LINKED ISSUE #38: highlight manager: adjust highlight on 'navigational-key' event

jaredhirsch:
Reopening previous PR, need to take a final look. @chuckharmston, if you want to verify this works for you, that would be great, but no worries if not.

Remainder of commit message follows:

---

Highlights:
- Always handle navigational keys, now that we are replacing the XBL
  focus logic.
- Use mutation observer to steal focus over and over as rows are
  inserted into the popup.
- Use 2 requestAnimationFrames to minimize the amount of blue highlight
  flickering on the zeroth result list item as results stream in.
- Take an initial swing at the highlight style.
- Improve documentation throughout.

Fixes #38.

### DIFF
--- a/chrome/content/binding.xml
+++ b/chrome/content/binding.xml
@@ -4,6 +4,13 @@
           xmlns:html="http://www.w3.org/1999/xhtml"
           xmlns:xul="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
           xmlns:xbl="http://www.mozilla.org/xbl">
+
+  <![CDATA[
+    /*
+     * Override the default binding, inserting a recommendation element in
+     * the popup above the list of history/suggestion results.
+     */
+  ]]>
   <binding id="recommendation-popup" extends="chrome://browser/content/urlbarBindings.xml#urlbar-rich-result-popup">
     <content ignorekeys="true" level="top" consumeoutsideclicks="never" aria-owns="richlistbox">
       <xul:hbox id="universal-search-recommendation" collapsed="true"/>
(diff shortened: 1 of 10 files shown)


jaredhirsch:
I need a break, then I'll do a final pass to make sure I addressed everything from the previous discussion.

lib/universal-search.js
```diff
@@ -183,7 +193,7 @@ Search.prototype = {
     const app = win.universalSearch;
 
     app.popup.destroy();
-    app.recommendationView.destroy();
+    app.recommendation.destroy();
```
jaredhirsch, line 196: this was a typo I missed a while ago

lib/universal-search.js
```diff
@@ -149,6 +156,17 @@ Search.prototype = {
...
     app.recommendationServer.init();
 
+    // NOTE: order matters here, the urlbar and highlightManager both need
+    // to be handed the initialized recommendation
+    app.recommendation = new app.Recommendation({
+      eTLD: Services.eTLD,
+      events: app.events,
+      io: Services.io,
+      win: win,
+      RecommendationRow: app.RecommendationRow
```
jaredhirsch, line 166: passing the constructor into the constructor, rather than fishing it out of `win.universalSearch`

lib/ui/recommendation.js
```diff
@@ -81,7 +101,7 @@ Recommendation.prototype = {
     if (this.el) {
       const container = this.win.document.getElementById(RECOMMENDATION_ID);
       this.timeout = this.win.setTimeout(function() {
-        // this.el.collapsed = true;
+        this.el.collapsed = true;
```
jaredhirsch, line 104: I think this was a typo that slipped through

jaredhirsch:
Hmm, after rebasing, looks like the recommendation isn't getting fully removed from the DOM. I'm out of time for tonight, but will fix tomorrow and merge.

screen shot 2016-03-10 at 7 01 03 pm

chuckharmston:
I'm not sure if it's the removal from the DOM or a positioning miscalculation; I think the "search with" row is being shown over the top of the recommendation, and only the URL shows through for...reasons.

Worth noting: I put `visibility: visible` on the URL, because it's hidden by default.

chuckharmston:
screen shot 2016-03-11 at 9 01 55 am

I can still get it to double-highlight by typing quickly.

chuckharmston:
CSS things for me to file and do when this merges:
- The label text should be some semi-transparent white, rather than grey. Will consult design for that.
- The image isn't showing up when highlighted, likely because it has `z-index: -1`, which is needed for its parent's inset box shadow to overlay it. When highlighted, we can set a normal z-index, unset the box shadow, and set a white background color to make sure transparent images look okay.

chuckharmston:
Documentation is fantastic. Thanks so much for that.

chrome/content/binding.xml
```diff
@@ -12,12 +19,98 @@
...
+    <implementation>
+      <![CDATA[
+        /*
+         * The selectedIndex setter sets the blue highlight on the currently-
+         * selected item in the results list. The existing implementation,
+         * defined in autocomplete.xml, automatically closes the popup if no
+         * results list items are selected (selectedIndex = -1).
+         *
+         * Override the setter, removing the auto-close behavior, so that the
+         * universal search recommendation element can steal the highlight.
```
chuckharmston, line 31: :boom:

lib/ui/recommendation.js
chuckharmston, line 104: Yeah, it was. If you're pinning the popup, this hides the recommendation after 50ms, which sort of makes the pin useless. I'll make sure to put this behind a conditional when working on #16

chuckharmston:
One more weird behavior: if a recommendation comes in while something is selected, the highlight is lost. It's very disorienting. 

You can see that here, represented by that highlight jumping around the 2 second markL https://chchc.ch/LjyWi

chrome/content/binding.xml
jaredhirsch, line 31: haha, nicely gif'd

jaredhirsch:
@chuckharmston That is indeed weird, mind filing as a followup bug? This has been brewing for too long, I think it's time to land it

PR #88, opened 2016-03-16
Manage mouse highlight (fixes #39)

### LINKED ISSUE #39: manage mouse highlight
Existing behavior:
- as you mouse over items in the popup, a mousemove listener ([autocomplete.xml, around line 2050](https://dxr.mozilla.org/mozilla-central/source/toolkit/content/widgets/autocomplete.xml#2051)) on the `richlistbox` sets `selectedIndex` of the popup to the hovered item.
- if you mouse out of the popup, the last hovered item stays highlighted

Integrating the recommendation into the existing hover behavior:
- when the user mouses into the recommendation node (not sure about mouseenter vs mousemove for this), highlight the recommendation, and remove the highlight from the richlistbox by setting `popup.selectedIndex = -1`.
- when the user mouses out of the recommendation node, we want to keep the recommendation highlighted if they are mousing out of the popup, but remove the highlight if they are mousing into the richlistbox. `mouseleave` events expose a `relatedTarget` property that might tell us what we need to know--or we just setTimeout, then check to see if the selectedIndex isn't -1.

I'm honestly not sure if this should be the responsibility of the highlight manager, or the recommendation view, or even the popup: the highlight manager is currently the only thing that knows about the popup's highlight state, but mouse events on the recommendation element seem like the concern of the recommendation view. Maybe the answer is to roll the highlight manager into the popup, which does basically nothing right now. Any solution that works, works; the current division of labor among the components is just a guess.

jaredhirsch:
@chuckharmston R?

The mouse highlight logic is collected in the HighlightManager's attachMouseListeners method, and should hopefully be fairly readable.

On a very happy side note, I discovered that leaving the `selectedIndex` state where it was originally (on the results list) lets us remove 95% of the XBL code introduced in the key focus PR! :tada: :fireworks:

This code might be slightly easier to read if you go through it commit by commit. I'll point out a few changes that fixed unrelated bugs/typos I noticed while in the code.

### DIFF
--- a/chrome/content/binding.xml
+++ b/chrome/content/binding.xml
@@ -19,85 +19,6 @@
         <children/>
       </xul:hbox>
     </content>
-    <implementation>
-      <![CDATA[
-        /*
-         * The selectedIndex setter sets the blue highlight on the currently-
-         * selected item in the results list. The existing implementation,
-         * defined in autocomplete.xml, automatically closes the popup if no
-         * results list items are selected (selectedIndex = -1).
-         *
-         * Override the setter, removing the auto-close behavior, so that the
-         * universal search recommendation element can steal the highlight.
-         */
-      ]]>
-      <field name="_selectedIndex">-1</field>
-      <property name="selectedIndex">
-        <getter>
-          return this._selectedIndex;
-        </getter>
-        <setter><![CDATA[
-          var resultsContainer = document.getAnonymousElementByAttribute(gURLBar.popup, 'anonid', 'richlistbox');
-
-          // NOTE: getElementsByClassName is needed to obtain a live NodeList.
-          // The results aren't inserted all at once: they are inserted a few
-          // at a time over many successive turns. In order to avoid setting
-          // the selectedIndex past the end of the results list, and also to
-          // avoid mistakenly concluding that the results list is empty, the
-          // NodeList needs to be live, not static.
-          var resultRows = resultsContainer.getElementsByClassName('autocomplete-richlistitem');
-
-          if (!resultRows || !resultRows.length) {
-            return;
-          }
-
-          if (val > resultRows.length - 1) {
-            val = resultRows.length - 1;
-          } else if (val < -1) {
-            val = -1;
-          }
-
-          // Unselect all the rows (in case of any weird multiple-focus bugs),
-          // then immediately select the appropriate row, to hopefully batch
-          // the selection changes into a single DOM update.
-          Array.prototype.forEach.call(resultRows, function(row, i) {
-            row.selected = false;
-          });
-
-          // Highlight the row, if it's found, else highlight nothing.
-          if (resultRows[val]) {
-            resultRows[val].selected = true;
-            // Scroll the row into view, if necessary.
-            resultsContainer.ensureIndexIsVisible(val);
-            this._selectedIndex = val;
-            return val;
-          } else {
-            this._selectedIndex = -1;
-            return -1;
-          }
-        ]]></setter>
-      </property>
-
-      <![CDATA[
-        /*
-         * The urlbar stops working if getNextIndex is not overridden here.
-         * This implementation doesn't handle certain edge cases, like
-         * scrolling off the top or bottom of the list, but whatever code
-         * calls this code doesn't seem to hit that bug.
-         */
-      ]]>
-      <method name="getNextIndex">
-        <parameter name="reverse"/>
-        <parameter name="amount"/>
-        <parameter name="index"/>
-        <parameter name="maxRow"/>
-        <body><![CDATA[
-          let realAmount = amount % maxRow;
-          this._selectedIndex = reverse ? this._selectedIndex - realAmount: this._selectedIndex + realAmount;
-          return this._selectedIndex;
-        ]]></body>
-      </method>
-    </implementation>
   </binding>
 
   <![CDATA[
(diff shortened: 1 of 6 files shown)


lib/ui/highlight-manager.js
```diff
@@ -32,31 +36,94 @@ HighlightManager.prototype = {
...
   destroy: function() {
     this.resultsObserver.disconnect();
+    this.detachMouseListeners();
+
     this.events.unsubscribe('recommendation-shown', this.stealHighlight);
     this.events.unsubscribe('navigational-key', this.adjustHighlight);
+
     delete this.popup;
-    delete this.recommendationView;
+    delete this.recommendation;
```
jaredhirsch, line 50: this was a typo

lib/ui/highlight-manager.js
```diff
@@ -102,20 +169,23 @@ HighlightManager.prototype = {
...
   clearHighlight: function() {
-    const resultsContainer = this.win.document.getAnonymousElementByAttribute(this.popup, 'anonid', 'richlistbox');
-    const resultRows = resultsContainer.getElementsByClassName('autocomplete-richlistitem');
-    Array.prototype.forEach.call(resultRows, row => { row.selected = false; });
+    this.popup.el.richlistbox.selectedIndex = -1;
+    this.recommendation.isHighlighted = false;
+  },
+  clearRecommendationHighlight: function() {
+    // See attachMouseListeners() for documentation.
+    this.recommendation.isHighlighted = false;
```
jaredhirsch, line 188: Seems funny to have a one-line method, but we need a non-anonymous wrapper so that we can unsubscribe the listener for the `'results-list-mouse-highlighted'` event.

lib/ui/popup.js
```diff
@@ -20,7 +27,33 @@ Popup.prototype = {
...
+
+    // Throttle mouseover events, which fire rapidly if the user quickly moves
+    // the mouse across different DOM elements in the results list.
+    if (this.mouseOverTimeout) {
+      return;
+    }
+    this.mouseOverTimeout = this.win.setTimeout(() => {
+      this.win.clearTimeout(this.mouseOverTimeout);
+      this.mouseOverTimeout = null;
+    }, 15);
```
jaredhirsch, line 55: I use a nearly-identical throttle syntax elsewhere in this PR. I think it's clearer in this form than wrapped in a throttle utility method, but open to alternative points of view :-)

lib/ui/recommendation.js
```diff
@@ -44,37 +48,71 @@ Recommendation.prototype = {
...
     this.events.unsubscribe('before-popup-hide', this.hide);
 
+    this.win.clearTimeout(this.mouseMoveTimeout);
+
     delete this.el;
     delete this.win;
   },
-
-  _pollForElement: function() {
+  _pollForElement: function(cb) {
```
jaredhirsch, line 87: Using a callback, not a promise, because we don't need the rejection option, and this is the least amount of code that works.

lib/universal-search.js
```diff
@@ -175,18 +175,19 @@ Search.prototype = {
...
       win: win,
       events: app.events
     });
     app.popup.init();
+
+    app.highlightManager = new app.HighlightManager({
+      win: win,
+      events: app.events,
+      recommendation: app.recommendation,
+      popup: app.popup
```
jaredhirsch, line 188: adjusting initialization order so we can pass the popup object into the highlightManager

chrome/content/binding.xml
```diff
@@ -116,4 +37,18 @@
...
+     * fired on it.
+     */
+  ]]>
+  <binding id="autocomplete-richlistbox" extends="chrome://global/content/bindings/autocomplete.xml#autocomplete-richlistbox">
+    <handlers>
+      <handler event="mouseover" phase="capturing"><![CDATA[
+        window.universalSearch.popup.onResultsMouseOver(event);
+      ]]></handler>
+    </handlers>
+  </binding>
```
chuckharmston, line 53: Wait. This is it?

lib/ui/highlight-manager.js
```diff
@@ -32,31 +36,94 @@ HighlightManager.prototype = {
...
     this.events.unsubscribe('navigational-key', this.adjustHighlight);
+
     delete this.popup;
-    delete this.recommendationView;
+    delete this.recommendation;
     delete this.win;
   },
+  attachMouseListeners: function() {
+    // This function is important because it gathers docs for all the mouse
+    // event handlers in a single place.
```
chuckharmston, line 55: Cackled.

lib/ui/popup.js
chuckharmston, line 55: Completely agree.

chuckharmston:
This is exceptional. Seriously great job. I was a little concerned about the spiraling complexity just to reach parity, but this eased all of my concerns.

I've got meetings, but will return with more review goodness.

chrome/content/binding.xml
jaredhirsch, line 53: heh, yup

lib/ui/highlight-manager.js
jaredhirsch, line 55: ^_^

jaredhirsch:
Thanks, Chuck!

chuckharmston:
I've spent quite a while playing with it, and the only thing I've noticed that didn't feel natural was that a number of times I tried to use `command` as a modifier, it didn't do what I expected. I haven't been able to reliably reproduce, so it may just be that my keyboard is sticky.

Worth monitoring, anyway.

jaredhirsch:
screen shot 2016-03-17 at 12 38 35 pm

/me merges

PR #90, opened 2016-03-17
Removes inset box-shadow from highlighted key image and logo (closes #84).

### LINKED ISSUE #84: Recommendation image not showing up while highlighted.
edd49a4e-e767-11e5-88fa-9bae2645a788

This is likely because it has z-index: -1, which is needed for its parent's inset box shadow to overlay it. 

Fix: when highlighted, we can set a normal z-index, unset the box shadow, and set a white background color on the image.

chuckharmston:
r? @6a68 
## Before

screen shot 2016-03-17 at 1 58 27 pm
## After

screen shot 2016-03-17 at 1 56 20 pm

screen shot 2016-03-17 at 1 56 24 pm

_NB: Google's image has a transparent background._

### DIFF
--- a/chrome/skin/style.css
+++ b/chrome/skin/style.css
@@ -155,3 +155,11 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
   color: HighlightText;
   opacity: 0.67;
 }
+#universal-search-recommendation.highlight #universal-search-recommendation-image {
+  background-color: #FFF;
+  box-shadow: 0;
+}
+#universal-search-recommendation.highlight #universal-search-recommendation-logo,
+#universal-search-recommendation.highlight #universal-search-recommendation-keyimage {
+  z-index: 1;
+}


chrome/skin/style.css
```diff
@@ -151,3 +151,12 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 #universal-search-recommendation.highlight #universal-search-recommendation-url {
   color: HighlightText;
 }
+#universal-search-recommendation.highlight #universal-search-recommendation-image {
+  background-color: #FFF;
+  box-shadow: 0;
```
jaredhirsch, line 160: What if we just remove the `box-shadow` at line 53? It looks like the same shadow is also applied to the letterbox on line 74, so it shouldn't hurt the letterbox case.

jaredhirsch, line 160: Oh, this is conditionally applied when the highlight is present, I see. Let me play with the branch to look for jank due to z-index recalculation, but this is probably fine.

chuckharmston, line 160: That's there to provide a bounding box for images with white or transparent backgrounds, e.g. 

screen shot 2016-03-17 at 2 06 17 pm

jaredhirsch:
@chuckharmston sure, let's do it, looks like it needs a rebase though

PR #91, opened 2016-03-19
Update urlbar state when the popup selection changes

### LINKED ISSUE #46: correctly update gURLBar state when recommendation is highlighted
When the recommendation gets highlighted, we need to put the URL in the urlbar.

However, if the user keys to an item in the popup, then hits Escape to dismiss the popup, the urlbar state has to be reverted. If we set state in the urlbar correctly, this behavior should be preserved.

There's a buggy implementation of this in the old addon; need to look carefully at `urlbarBindings.xml`, and figure out which of these bits of state is relevant:
- `gURLBar.value`
- `gURLBar._value`
- `gURLBar.inputField.value`
- `gURLBar.valueIsTyped`
- `gURLBar.valueIsPasted`
- `gBrowser.userTypedValue`
- `gBrowser.userTypedClear`

### LINKED ISSUE #47: implement urlbar navigate method
This should be an easy one: if the user wants to navigate to the recommendation URL, we need a method on the urlbar that navigates to the urlbar contents.

I think `gURLBar.handleCommand(url)` should do it, but there are no doubt some funny edge cases.

jaredhirsch:
@chuckharmston R?

Thankfully, the gURLBar.textValue setter lets us hook into an enormously
complex system with a simple one-liner.

Fixes #46, fixes #47.

### DIFF
--- a/lib/ui/highlight-manager.js
+++ b/lib/ui/highlight-manager.js
@@ -208,6 +208,9 @@ HighlightManager.prototype = {
     const selectedIndex = this.popup.el.selectedIndex;
     const recommendationVisible = this.recommendation.el && !this.recommendation.el.collapsed;
     const recommendationHighlighted = this.recommendation.el && this.recommendation.isHighlighted;
+    // To update the urlbar contents quickly, we'll keep track of the new
+    // index that's gaining focus, grab its url, and pass it to the urlbar.
+    let newIndex;
 
     // Clear all highlights. The DOM is now dirty and not trustworthy.
     this.clearHighlight();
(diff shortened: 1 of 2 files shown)


lib/ui/highlight-manager.js
```diff
@@ -252,16 +257,35 @@ HighlightManager.prototype = {
...
+      newIndex = evt.forward ? selectedIndex + 1 : selectedIndex - 1;
+      this.popup.el.selectedIndex = newIndex;
+    }
+
+    // Finally, figure out the new url, and tell the urlbar that the selection
+    // has changed.
+    let newUrl;
+    if (newIndex === -1) {
+      let urlEl = this.recommendation.el.querySelector('#universal-search-recommendation-url');
+      newUrl = urlEl && urlEl.textContent;
```
chuckharmston: Rather than use `textContent`, can we use a `data-` attribute for meaningful data?

chuckharmston:
This is great work. Feel free to merge at will!

lib/ui/highlight-manager.js
jaredhirsch: :+1:

jaredhirsch:
Thanks for taking a look! Merging

PR #94, opened 2016-03-20
Creates separate stylesheets for each platform (closes #78).

### LINKED ISSUE #78: Use separate highlight styles for different platforms
_This might be a little too ambitious for a good first bug, so if you are a new contributor and are interested in taking this on, please ping me to get guidance before you start. IRC is best (my nick is _6a68), or leave a comment in this bug._

Right now, we style the recommendation [the same way](https://github.com/mozilla/universal-search/blob/master/chrome/skin/style.css) for all platforms. In general, this is fine, but it would be nice to have platform-specific highlight behavior when the recommendation row is selected.

Here's a sketch of the tasks required:
- Create individual files in the `chrome/skin/` directory for mac, windows, and linux.
- In each file, copy over the highlight style for each platform (see [this comment](https://github.com/mozilla/universal-search/pull/73#discussion_r55405801) for links to files). You want to take the `.autocomplete-richlistitem[selected=true]` rules, but apply them to the `#universal-search-recommendation` selector.
- Modify the [stylesheet loading code](https://github.com/mozilla/universal-search/blob/master/lib/universal-search.js#L82-L94) to add a check for the platform type, then load a second, platform-specific highlight stylesheet.
- You can get the platform type by checking the value of `Services.appinfo.OS`. The possible values of interest are `'Darwin'`, '`Linux'`, and `'WINNT'`. (It takes about 4 hops in MDN to track this info down, directly inserting the answer here to keep the bug manageable overall.)
- Once the changes are made, restart Firefox and see if the recommendation highlight matches the highlight of the other rows in the autocomplete dropdown. If it looks good, open a pull request. Don't worry about checking all three platforms; I can check that.

kaganjd:


chuckharmston:
In general, it'd be fantastic to cut out the duplication between each of these files. Perhaps you could keep all the common styles in a single file and [`@import`](https://developer.mozilla.org/en-US/docs/Web/CSS/@import) it into each platform-specific one?

lib/universal-search.js
```diff
@@ -88,14 +88,27 @@ Search.prototype = {
   _loadStyleSheets: function(win) {
     const doc = win.document;
     const docEl = doc.documentElement;
+    //check whether platform type is Darwin, Linux, or WINNT
```
chuckharmston: We try to make comments complete sentences that start with a capital letter and end with punctuation.

chrome/skin/style-osx.css
```diff
@@ -0,0 +1,174 @@
...
+textbox#urlbar {
+  -moz-binding: url("chrome://universalsearch/content/binding.xml#recommendation-urlbar");
+}
+
+popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
+  -moz-binding: url("chrome://universalsearch/content/binding.xml#autocomplete-richlistbox");
+}
+
+#universal-search-recommendation {
+  /*made these obvious colors to test whether stylesheet was loading*/
```
chuckharmston: No need to keep this around.

chuckharmston:
Would you mind squashing the two commits? An [interactive rebase](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history) should help you with that.

chuckharmston:
Great start! Feel free to ping me once you've addressed the commits above or with any questions.

Thanks for the contribution!

chrome/skin/style-osx.css
kaganjd: When I delete 13-26, the styling disappears, but I'm deleting 28-31 (the test styling). Does that work?

chuckharmston: Oh, yes, sorry for ambiguity. I was specifically talking about the test styling.

jaredhirsch:
@kaganjd sorry about the conflicting reviewer feedback, let's close this one and finish out the changes in #95. Also, note that the patch doesn't have to land before Wednesday: you just need your application in by then. Mentioning this in case you have limited time over the next day or two.

PR #97, opened 2016-03-22
Update devprefs add-on url at README

anarute:


chuckharmston:
Looks like an overzealous find/replace job hit this URL by mistake. Thanks, @anarute!

PR #101, opened 2016-03-23
Respects browser.urlbar.maxCharsForSearchSuggestions for recommendations (closes #13).

### LINKED ISSUE #13: hide recommendation node if the user's typed string is too long
The existing popup only sends strings of up to 20 characters to the search suggestion service.

We should do something similar by modifying the `search` method on the [`RecommendationServer`](https://github.com/mozilla/universal-search/blob/master/lib/recommendation-server.js) to check the length of `payload.query`, and do nothing if the length is > 20.

chuckharmston:
r? @6a68

### DIFF
--- a/lib/recommendation-server.js
+++ b/lib/recommendation-server.js
@@ -27,6 +27,7 @@ function RecommendationServer(opts) {
   this.xhr = opts.xhr;
   this.console = opts.console;
   this.prefs = opts.prefs;
+  this.urlbarPrefs = opts.urlbarPrefs;
   this.server = this.prefs.getPrefType('server') ?
                 this.prefs.getCharPref('server') :
                 'http://universal-search.dev';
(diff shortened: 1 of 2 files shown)


lib/recommendation-server.js
```diff
@@ -42,6 +43,12 @@ RecommendationServer.prototype = {
 
     this.events.subscribe('urlbar-change', this.search);
   },
+  maxCharsForRecommendation: function() {
+    const name = 'maxCharsForSearchSuggestions';
+    return this.urlbarPrefs.getPrefType(name) ?
+           this.urlbarPrefs.getIntPref(name) :
+           20;
+  },
```
jaredhirsch, line 52: This looks great, but I would only run this pref check once, inside the init() method, and set the result on the object. We call the search function a lot, and I'm not sure if hitting the pref system would slow things down.

Because this initialization happens for each new window, I think we're fine to take this shortcut in a prototype. If we wanted to be exactly correct, we would additionally set up a preferences observer, but let's skip that.

jaredhirsch:
@chuckharmston I see Gecko is soaking into your brain, nicely done. One suggested change mentioned inline, feel free to merge when that's done.

lib/recommendation-server.js
chuckharmston, line 52: Yeah, the correct answer is definitely to use a preferences observer, but I think #58 is pretty far down the priority list. I figured that respecting the preference should, in theory, take precedent over any trivial speed gains, but if we're only going to have a few thousand users it's just not a big deal.

PR #108, opened 2016-03-24
Prevents rerendering of duplicate recommendation (closes #106).

### LINKED ISSUE #106: Don't update recommendation when search changes if the recommendation doesn't change
We currently swap out the recommendation DOM object whenever a new recommendation is received. This results in a perceptible flicker, e.g. https://chchc.ch/MDjyqD 

We should conditionalize this to only update if the recommendation itself changes. JavaScript object equality is pretty complicated, but I think that `old.result.url === new.result.url && Object.keys(old.enhancements) === Object.keys(new.enhancements)` would be good enough for these purposes.

chuckharmston:
Before: https://chchc.ch/MDjyqD
After: https://chchc.ch/0F19xr

r? @6a68

### DIFF
--- a/lib/ui/recommendation.js
+++ b/lib/ui/recommendation.js
@@ -13,10 +13,13 @@ function Recommendation(opts) {
   this.events = opts.events;
   this.eTLD = opts.eTLD;
   this.io = opts.io;
+  this.objectUtils = opts.objectUtils;
+  this.RecommendationRow = opts.RecommendationRow;
+
   this.el = null;
   this.timeout = null;
-  this.RecommendationRow = opts.RecommendationRow;
   this.mouseMoveTimeout = null;
+  this.prev = null;
 
   Object.defineProperty(this, 'isHighlighted', {
     get: () => {
@@ -118,6 +121,16 @@ Recommendation.prototype = {
       this.win.clearTimeout(this.timeout);
       this.timeout = null;
     }
+
+    // Don't re-render the recommendation if the content hasn't changed since
+    // the previous recommendation.
+    if (this.objectUtils.deepEqual(data, this.prev)) {
+      this.el.collapsed = false;
+      this.events.publish('recommendation-shown', this);
+      return;
+    }
+
+    this.prev = data;
     const recommendation = new this.RecommendationRow(data, {
       eTLD: this.eTLD,
       io: this.io,
--- a/lib/universal-search.js
+++ b/lib/universal-search.js
@@ -9,6 +9,8 @@ XPCOMUtils.defineLazyModuleGetter(this, 'console',
   'resource://gre/modules/Console.jsm');
 XPCOMUtils.defineLazyModuleGetter(this, 'CustomizableUI',
   'resource:///modules/CustomizableUI.jsm');
+XPCOMUtils.defineLazyModuleGetter(this, 'ObjectUtils',
+  'resource://gre/modules/ObjectUtils.jsm');
 XPCOMUtils.defineLazyModuleGetter(this, 'PrivateBrowsingUtils',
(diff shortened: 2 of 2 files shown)


lib/recommendation-server.js
```diff
@@ -31,6 +31,8 @@ function RecommendationServer(opts) {
                 this.prefs.getCharPref('server') :
                 'http://universal-search.dev';
   this.search = this.search.bind(this);
+  this.prev = null;
+  this._resultHasChanged.bind(this);
```
jaredhirsch: Here, we technically want to assign the result of binding to the object, so that we can both attach and detach the same bound listener:

`this._resultHasChanged = this._resultHasChanged.bind(this)`

This is why you ran into some binding problems with `this` below. JS is weird, eh? :-)

lib/recommendation-server.js
```diff
@@ -55,6 +57,39 @@ RecommendationServer.prototype = {
     const endpoint = useDummyEndpoint ? 'dummy' : '';
     return `${this.server}/${endpoint}?q=${encodeURIComponent(query)}`;
   },
+
+  _resultHasChanged(data) {
+    /*
```
jaredhirsch: Took some serious digging in dxr, but it looks like you can load `ObjectUtils.jsm` and call `ObjectUtils.deepEqual()` to compare two objects. That should simplify this function quite a bit :-)

lib/recommendation-server.js
```diff
@@ -71,14 +106,21 @@ RecommendationServer.prototype = {
     }
   },
   search: function(payload) {
+    let server = this;
```
6a68: The binding issues here are just due to not assigning the bound `_resultHasChanged` back onto the object (see my comment at line 35). Once you make that change, you can get rid of the `server = this` assignment and update the code below. The arrow func preserves `this` binding, so it should just work. I'm happy to help debug if it continues to fight you :-)

lib/recommendation-server.js
```diff
@@ -71,14 +106,21 @@ RecommendationServer.prototype = {
...
     if (!payload.query) {
       return;
     }
-    this._search(payload.query).then(data => {
-      this.events.publish('recommendation', data);
+    server._search(payload.query).then(data => {
+      if (server._resultHasChanged(data)) {
```
jaredhirsch: Hmm. Now that I'm staring at this, it might be easier to just inline the object equality check?

`if (this.objectUtils.deepEqual(this.prev, data)) { ...`

jaredhirsch:
@chuckharmston Doh. I broke my rule about reading through twice before commenting, and started with not super useful low-level feedback.

After looking all the way through the code, I think it might be simpler if you did the duplicate check in the view's `show` method, not in the server lib:
- unset the timeout
- if the data is NOT a duplicate of `this.data`, create a new row and append it to the DOM, endif
- set collapsed to false and fire the recommendation-shown event (whether or not it was a dupe)

This would let us avoid introducing a new pubsub event (unless you want a new event for metrics?), and avoid introducing the `keep` method.

Does this make sense?

Removing a couple of now irrelevant inline comments you might have gotten emailed to you, sorry!

lib/recommendation-server.js
chuckharmston: It's weird, but I'm also dumb :)

chuckharmston:
I got hung up on the philosophical best place to put the logic, which is probably in the `RecommendationServer` singleton, but you have to bend over backward to make it work.

This is infinitely better, great calls all around. re-r? @6a68

jaredhirsch:
Looks great, and ObjectUtils.deepEqual plays nicely with nulls (you never know when Gecko JSMs will throw). Nice work @chuckharmston :-)

PR #111, opened 2016-03-25
Adds new staging server hostname; defaults to that.

chuckharmston:
r? @6a68


chuckharmston:
screen shot 2016-03-24 at 10 42 19 pm

PR #116, opened 2016-03-26
When the user types, highlight the urlbar, not the recommendation

### LINKED ISSUE #96: Recommendation shouldn't be highlighted until user arrows down
The default behavior highlights a row that corresponds to what would happen anyway if you hit Enter with the focus in the urlbar:

screen shot 2016-03-21 at 8 48 32 am

When we steal focus for the recommendation, it seems like hitting Enter would take you to the recommendation URL, but you actually have to key down, then back up, to get the recommendation URL into the urlbar (or click on it).
1. user types 'f'
   screen shot 2016-03-21 at 8 49 48 am
2. user keys down
   screen shot 2016-03-21 at 8 50 04 am
3. user keys back up to the recommendation, now the URL is in the urlbar
   screen shot 2016-03-21 at 8 50 08 am

Maybe the better interaction would be to not highlight the recommendation until the user hits the down arrow for the first time?

### LINKED ISSUE #112: When popup contains suggestions, keying backwards is broken
If the popup is filled with suggestions, not history results, it seems like keying "backwards" (up arrow key or shift-tab) from the urlbar causes the highlight to disappear.

Maybe the problem is that the current highlight manager approach is to check the number of visible rows at the _start_ of the calculation, but the suggestion results stream in relatively late, so the results length may be inaccurate.

If this is indeed the cause of the bug, the underlying cause is that we don't ensure that setting the highlight on an index actually results in highlighting a row. We could fix this by adding a method that synchronously checks the DOM, and highlights whatever the last row is.

jaredhirsch:
@chuckharmston R? Modified the highlight manager logic to put the focus in the urlbar (i.e., not highlight the recommendation or the results list) whenever the recommendation is visible. I wound up refactoring the highlight management logic, because it was a bit hard to puzzle through the cases as written. The [new version of the `HighlightManager.adjustHighlight` function](https://github.com/6a68/universal-search/blob/f407fa91db6cf9f5269d4fe43bf64596b3b35919/lib/ui/highlight-manager.js#L194) might be more readable than the unified diff. I rarely say this, but I'm actually pleased with how the code is shaping up ^_^

Desired behavior after applying this patch: when you type in the urlbar, if the recommendation is visible, neither the recommendation nor any results list item should have focus. The urlbar should appear to have focus.

The underlying reason for this change is that the highlighted thing should tell you what happens if you hit 'Enter', and we don't put the recommendation URL in the urlbar until you select it. In contrast, if the recommendation is not shown, the top results list item should be highlighted, because what it says (either 'search with &lt; search provider >' or 'visit &lt; url >' is actually what happens if you hit enter.

The rest of the commit message follows:
- Reorganize the highlight transition logic to be easier to read.
- The popup stores unused rows for later reuse, so only use visible
  rows for highlight transition calculations (fixes #112).

Fixes #96.

### DIFF
--- a/lib/ui/highlight-manager.js
+++ b/lib/ui/highlight-manager.js
@@ -139,15 +139,21 @@ HighlightManager.prototype = {
     }
   },
   stealHighlight: function() {
-    // Retake control of the highlight under two circumstances:
+    // If the recommendation is shown, the default behavior is for the urlbar
+    // to have the highlight. That is, neither the recommendation nor any items
+    // in the results list should be highlighted.
+    //
+    // We need to retake control of the highlight under two circumstances:
     // 1. When the recommendation has just been shown, if the user isn't keying
-    //   through the list already, then move the highlight from the top results
-    //   list item to the recommendation.
-    // 2. When a few results have been inserted into the DOM, causing the Gecko
-    //   code to move the highlight, steal it back.
+    //    through the list already, then move the highlight from the top results
+    //    list item to the urlbar.
+    // 2. After (1), when a few more results have been inserted into the DOM,
+    //    the Gecko code will take the highlight again; steal it back.
     //
     // The existing XBL code batches insertions in timeouts. By default,
-    // 30 rows are inserted in groups of 6, invoking setTimeout 5 times.
+    // the code invokes setTimeout 5 times, inserting 6 rows per turn, up
+    // to the default max of 30 rows.
+    //
     // Because the event loop already has up to 5 timers waiting, we can't
     // rely on setTimeout to steal the highlight: our request will be queued
     // behind all those other timers, causing a noticeably long flicker of the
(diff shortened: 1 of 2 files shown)


lib/ui/urlbar.js
```diff
@@ -33,9 +33,10 @@ Urlbar.prototype = {
     delete this.el;
     delete this.win;
   },
+  // Set the URL in the urlbar to either data.newUrl, if it's defined;
+  // otherwise, reset the URL to the actual value the user typed.
```
chuckharmston, line 37: Grammar police :oncoming_police_car:: "either" should always be used with "or".

> Set the URL in the urlbar data.newUrl, if defined; otherwise reset the URL to the actual value the user typed.

chuckharmston:
I'm not sure the UX here is right; if we're going to prevent the top result from being highlighted by default if it's a recommendation, it seems like we should never highlight the top result. It's very disorienting to be typing and have the highlight flash in while typing a string: https://chchc.ch/PgsBdx

lib/ui/highlight-manager.js
```diff
@@ -155,27 +161,25 @@ HighlightManager.prototype = {
...
+    this.clearHighlight();
     this.win.requestAnimationFrame(() => {
-      this.recommendation.isHighlighted = true;
-      this.popup.el.selectedIndex = -1;
+      this.clearHighlight();
       this.win.requestAnimationFrame(() => {
-        this.recommendation.isHighlighted = true;
-        this.popup.el.selectedIndex = -1;
+        this.clearHighlight();
       });
```
chuckharmston, line 183: This is still the biggest
xait3izg

lib/ui/highlight-manager.js
```diff
@@ -199,93 +203,133 @@ HighlightManager.prototype = {
...
+    let shouldHighlightRecommendation;
 
     // Clear all highlights. The DOM is now dirty and not trustworthy.
     this.clearHighlight();
 
+    // Now, we look at the state of the recommendation and the results list,
+    // and decide what should be highlighted next.
+
+    // These utility functions abstract out the state changes, and make the
+    // state transition logic more readable.
```
chuckharmston, line 232: This makes the code really grokkable, very appreciated.

chuckharmston:
Code is great, logic is fantastic. Just not quite sure we're there on the UX yet, as noted above.

jaredhirsch:
@chuckharmston Yeah, I hear you on the UX concerns. Discussed this with @bryanbell, he was fine with it. What would you say we land this (once I address other concerns in the review), then file a needs-UX followup bug?

> I'm not sure the UX here is right; if we're going to prevent the top result from being highlighted by default if it's a recommendation, it seems like we should never highlight the top result. It's very disorienting to be typing and have the highlight flash in while typing a string: https://chchc.ch/PgsBdx

lib/ui/highlight-manager.js
jaredhirsch, line 232: :tada: Glad it seems easier to navigate!

lib/ui/highlight-manager.js
jaredhirsch, line 183: Heh. You mean the madness of rAF and setTimeout in a Godzilla vs. Mothra-like battle for dominance of the frame repaint?

lib/ui/urlbar.js
jaredhirsch, line 37: Cool, fixed.

chuckharmston:
File that bug and ship it!

jaredhirsch:
:beers: Filed https://github.com/mozilla/universal-search/issues/118. Merging.

PR #128, opened 2016-04-04
Show recommendations in response to 'paste' events on the urlbar

### LINKED ISSUE #68: fire 'urlbar-change' event on urlbar paste or cut events
Figure out how to detect and handle clipboard events that modify the contents of the urlbar.

jaredhirsch:
@chuckharmston r?

Here's the rest of the commit message:

Note: trying to listen for 'cut' events on the urlbar didn't work at
all. It seems fairly unusual for part of the urlbar contents to be cut,
so rather than look for a comprehensive fix, just deciding to stick with
the simple, much more common case of paste support.

Fixes #68.

### DIFF
--- a/lib/ui/urlbar.js
+++ b/lib/ui/urlbar.js
@@ -16,6 +16,7 @@ function Urlbar(opts) {
   this.recommendation = opts.recommendation;
 
   this.navigate = this.navigate.bind(this);
+  this.publishUrlbarUpdate = this.publishUrlbarUpdate.bind(this);
   this.updateTextValue = this.updateTextValue.bind(this);
 }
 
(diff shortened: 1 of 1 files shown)


chuckharmston:
Huh, I am surprised that this works. When I took a shot at this bug, I couldn't find any documentation showing that `onPaste` was supported by XUL, instead going down a deep rabbit hole with [`onInput`](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/oninput) and onKeypress. Excited to try it out!

chuckharmston:
Unfortunately, I'm just not seeing the paste triggering a state change. [See this](https://chchc.ch/lIaBP8), where a paste does not even send a new request to the server. I did a little bit of digging to see if I could figure out why—my initial idea was that something funny was happening with the timeout—but I couldn't even see a log from inside of `publishUrlbarUpdate`.

jaredhirsch:
@chuckharmston Hmm, I'm not sure what was up with this branch. I just rebased and force-pushed, and it seems to be working for me locally. Mind giving it another try?

<iframe width="854" height="480" src="https://www.youtube.com/embed/0znqX8csr1o" frameborder="0" allowfullscreen></iframe>

jaredhirsch:
Here's a slightly too short video of pasting a value and seeing a recommendation - it happens as soon as the video starts https://www.youtube.com/watch?v=0znqX8csr1o

chuckharmston:
`¯\_(ツ)_/¯` works now!

PR #129, opened 2016-04-05
Allow the popup to be opened by up or down arrows

### LINKED ISSUE #113: Up or Down arrow keys in the urlbar should open the popup
Our add-on doesn't currently reopen the popup if a down or up arrow key is pressed while the urlbar is focused.

Note that this should NOT be the case for tab / shift-tab, because tabbing is used to traverse the browser window.

jaredhirsch:
Fixes #113.

@chuckharmston r? You can test this by hitting the up or down arrows when the urlbar is focused, but the popup is not open.


chuckharmston:
:ship:

PR #132, opened 2016-04-05
Simplify recommendation-row check for keys in server response

### LINKED ISSUE #87: Broken null check in recommendation row
[This line](https://github.com/mozilla/universal-search/blob/d7d38e0319890624a2d07a3450be8086906542eb/lib/ui/recommendation-row.js#L187) fails to distinguish between null and empty objects, because `typeof null` is `'object'`.

jaredhirsch:
Fixes #87 

@chuckharmston R?

I found my way to this line in the course of debugging a few weeks ago, but I don't know what the symptom was.

At any rate, the `_get` method is called in a lot of places, and it looked pretty formidable as written. Adjusting to use a simpler approach.

### DIFF
--- a/lib/ui/recommendation-row.js
+++ b/lib/ui/recommendation-row.js
@@ -180,11 +180,11 @@ RecommendationRow.prototype = {
     Attempt to return a value from the recommendation data at the passed key
     string. If not set, falsy, or an empty object, return null;
     */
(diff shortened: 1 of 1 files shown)


jaredhirsch:
meh, this is arguably not valuable, and I'm seeing some bugs in my version, so, closing.

PR #133, opened 2016-04-06
Adds server log format example to metrics docs.

chuckharmston:
r? @clouserw @nchapman @6a68 @mostlygeek 

Read nicely here: https://github.com/chuckharmston/universal-search/blob/server-log-format/docs/metrics.md#on-the-server

For this first pass, I've chosen a subset of the [standard fields](https://github.com/mozilla-services/Dockerflow/blob/master/docs/mozlog.md#application-request-summary-type-requestsummary) for the `request.summary` type that seem useful for us. Happy to take feedback on missing ones that might be useful, or those that are unnecessary.

Companion pull request for the server coming soon (for https://github.com/mozilla/universal-search-recommendation/issues/35).

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -66,6 +66,73 @@ Additionally, we will discard outliers on a schedule:
 - Any queries that have only been made once.
 - Any queries longer than 6 characters that have been made less than 10% as frequently as the most common query of that length (e.g. if the most frequently-made 8-character query has been made 1000 times, all 8-character queries made fewer than 100 times will be discarded).
 
+The logs will be printed to stdout in the standard [mozlog format](https://github.com/mozilla-services/Dockerflow/blob/master/docs/mozlog.md):
+
+```json
+{
+    "EnvVersion": "2.0",
+    "Hostname": "socket.gethostname()",
+    "Logger": "universalSearch",
+    "Pid": 9,
+    "Severity": 6,
+    "Timestamp": 1459980510934186496,
+    "Type": "request.summary",
+
+    "Fields": {
+        "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:48.0) Gecko/20100101 Firefox/48.0",
+        "errno": 0,
+        "lang": "en-US,en;q=0.5",
+        "method": "GET",
+        "path": "/",
+        "t": 35,
+
+        "classifiers": [
+            "keyimage",
+            "favicon"
+        ],
+        "predicates": {
+            "is_protocol": false,
+            "query_length": false,
+            "is_hostname": false
+        },
+        "query": "the mart",
+        "status_code": 200
+    }
+}
+```
+
+
+#### Standard `mozlog` data
+
+* `EnvVersion`: `"2.0"`, static.
+* `Hostname`: the hostname, as reported by Python's `socket.gethostname()` function.
+* `Logger`: `"universalSearch"`, static.
+* `Pid`: the process number.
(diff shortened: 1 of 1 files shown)


mostlygeek:
LGTM!

docs/metrics.md
```diff
@@ -66,6 +66,73 @@ Additionally, we will discard outliers on a schedule:
...
+
+#### Custom fields
+
+* `Fields.classifiers`: an array of classifiers applied to the result. Not collected if any predicates are `true`.
+* `Fields.predicates`: an object of predicates to further data collection; if any of these are `true`, no query data is recorded.
+    * `Fields.predicates.is_hostname`: `true` if the query appears to be a hostname.
+    * `Fields.predicates.is_protocol`: `true` if the query appears to begin with a protocol.
+    * `Fields.predicates.query_length`: `true` if the query is longer than 20 characters.
+* `Fields.query`: the value of the `q` querystring parameter. Not collected if any predicates are `true`.
+* `Fields.status_code`: the HTTP status code of the response. Not collected if any predicates are `true`.
```
ghost, line 134: Isn't this already in Fields.errno above?

chuckharmston, line 134: Not quite. `errno` should be `0` if not an error, but we get valuable information from the differences between `200`, `202`, `304`, etc.

docs/metrics.md
```diff
@@ -66,6 +66,73 @@ Additionally, we will discard outliers on a schedule:
...
+* `Fields.t`: the number of ms elapsed in the processing of the request.
+
+
+#### Custom fields
+
+* `Fields.classifiers`: an array of classifiers applied to the result. Not collected if any predicates are `true`.
+* `Fields.predicates`: an object of predicates to further data collection; if any of these are `true`, no query data is recorded.
+    * `Fields.predicates.is_hostname`: `true` if the query appears to be a hostname.
+    * `Fields.predicates.is_protocol`: `true` if the query appears to begin with a protocol.
+    * `Fields.predicates.query_length`: `true` if the query is longer than 20 characters.
```
ghost, line 132: We have query below -- seems a little strange to have this here.  Is your plan to decide if something should be recorded on the client and then send it to the server where the server looks at what the client said and then does it?  If we have to write a stats handler with logic that discards rows, why not put the is_hostname/is_protocol/query_length logic there and just leave out the whole Fields.predicates from the logger?

chuckharmston, line 132: The thought here was that it's useful to know what percentage of queries are not being logged due to one of these. In the future we may have predicates that aren't just derived from the query—maybe we don't run a recommendation for IP addresses in some countries, for example—so I figured it'd be useful to log them so we always had all data. If there's a cost to logging too much, it's a trivial change.

chuckharmston, line 132: Code for this is here: https://github.com/mozilla/universal-search-recommendation/pull/91/files#diff-0fd92c74bc602e7689d84fec987158fbR20

docs/metrics.md
ghost, line 134: Yeah, I was going to ask about why errno was set to 0 instead of just keeping the value.  That's not your code?

chuckharmston, line 134: That's part of the standard `request.summary` log format that I separated out; docs here: https://github.com/mozilla-services/Dockerflow/blob/master/docs/mozlog.md

docs/metrics.md
ghost, line 132: Ah, I see what you're doing.  Seems like it will work

docs/metrics.md
ghost, line 134: Weird.

ghost:
r+ thanks

jaredhirsch:
Looks great. Thanks @chuckharmston! Merging

PR #134, opened 2016-04-07
Disable select and reset cursor on recommendation label (closes #99)

### LINKED ISSUE #99: Disable text selection on 'Recommended' text
If you mouse over the text that says 'Recommended', the pointer changes to a selection caret. We really just want the pointer to stay as the default/cursor style.

Applying the [`-moz-user-select`](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) style should solve the problem.

chuckharmston:
r? @6a68


jaredhirsch:
:tada: looks great, merging

PR #136, opened 2016-04-08
Hide recommendation if it duplicates an existing result

### LINKED ISSUE #12: handle duplicates between recommendation and rich list / urlbar autocomplete
I think we need some input from product here. Is the best thing to just not show the recommendation, if it duplicates one of the top 5 results (hence, is already visible in the popup)?

jaredhirsch:
@chuckharmston R?

This PR hides the recommendation when its domain duplicates the base domain of one of the results visible when the popup is first shown (up to 6 rows can be shown by default).

Unfortunately, this doesn't work for the first letter that triggers a duplicate, because of delays in the results rendering (the recommendation often seems to win the rendering race, and quickly re-hiding it looks super janky), but it does work for the second letter. I'll file a followup bug to try to hide the duplicate immediately, but it's been difficult so far (tried grabbing URLs from the C++ controller, but that didn't work). Maybe this will be a good enough duplicate trimming experience for the initial round of prototype testing.

Fixes #12.

### DIFF
--- a/lib/ui/highlight-manager.js
+++ b/lib/ui/highlight-manager.js
@@ -114,6 +114,9 @@ HighlightManager.prototype = {
     }
   },
   initMutationObserver: function() {
+    // TODO: Let's move the MutationObserver code to the popup, exposing the
+    // updates to the rest of the app via pubsub events. (#135)
+    //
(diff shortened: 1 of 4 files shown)


chuckharmston:
Looks and works awesome. :shipit:

Issue #143, opened 2016-04-18
Highlight stealing fails if the popup opens under the mouse


jaredhirsch:
If the popup opens under the mouse, one **or more** `mouseover` events may be fired. These events are used by the `HighlightManager` to detect when the user is mousing through the popup, in which case the highlight is not stolen. So, the net effect is that the highlight doesn't get stolen, but the user wasn't mousing through the popup. Awkward.

The right fix might be to listen for `mousemove` instead, as it's not fired when the popup renders under the mouse, only when the mouse pointer moves above a DOM element. My notes from the initial highlight stealing bug say that `mousemove` didn't work, but `mouseover` did, but I didn't write down the evidence that led me to that conclusion. Seems worth revisiting.

Issue #140, opened 2016-04-15
Steal highlight when only heuristic top result is shown


jaredhirsch:
For some reason, the existing highlight stealing code doesn't work in the case where the only thing shown in the popup is the 'heuristic' result:

<img width="485" alt="screen shot 2016-04-15 at 11 40 22 am" src="https://cloud.githubusercontent.com/assets/96396/14574311/b63c34c0-030f-11e6-808f-f4df79ce8357.png">

Figure out what's happening in this case, and steal the highlight to keep the experience consistent.

Not sure if this should block public launch or not, flagging product for feedback. @nchapman thoughts?

jaredhirsch:
Asked @nchapman in today's search planning meeting, answer was that this would be nice to have for public launch, but shouldn't block.

jaredhirsch:
I suspect the issue here is that `onResultsAdded` steals the highlight if `selectedIndex == -1`. See the `onResultsAdded` method in urlbarBindings on dxr: https://dxr.mozilla.org/mozilla-central/source/browser/base/content/urlbarBindings.xml#1614-1616

I wonder if we'd fix this bug by having the HighlightManager set `selectedIndex` to -2 instead when we are stealing the highlight.

Issue #118, opened 2016-03-28
Always hide the highlight when the popup first opens


jaredhirsch:
@chuckharmston raised this concern in the pull request that removed highlighting from the recommendation:

> I'm not sure the UX here is right; if we're going to prevent the top result from being highlighted by default if it's a recommendation, it seems like we should never highlight the top result. It's very disorienting to be typing and have the highlight flash in while typing a string: https://chchc.ch/PgsBdx

@nchapman / @bryanbell, thoughts?

I don't see a lot of options here:
- We could leave it like it is.
- We could try to strip the highlight out of the 'heuristic' ("search with...") top result. I'm not sure how much work that would be, probably a 'medium' 2-3 day bug.
- Note that it would be hard to remove the top 'heuristic' result completely, that's not an option here.

chuckharmston:
We could also revert to the previous behavior of having the recommendation be highlighted, right?

jaredhirsch:
@chuckharmston Actually, the previous behavior was broken: when you started typing, the recommendation was highlighted, but if you immediately hit Enter, you wouldn't be taken to the recommendation URL. You'd be taken to the URL/search page for whatever you had typed in the urlbar.

With the old system, the only way to surf to the recommendation URL would be to down arrow past it, then up arrow back to it, which would fill the urlbar with the recommendation URL, and _then_ hit Enter. It looked better, but the interaction was broken.

The basic principle is that the highlight should tell you what's about to happen when you hit Enter. Another basic assumption is that we don't want to interfere with the user typing something and hitting Enter without interacting with the popup in any way: that means we can't just auto-send them to the recommendation URL if they type a few chars, then hit Enter.

chuckharmston:
I vote for any option that maintains consistency for queries that do and do not have recommendations available. I guess that, in this case, that means that we should not highlight anything until the user interacts with it.

nchapman:
Yes I think we should hide the highlight so that it's consistent with the experience when there is a recommendation.

jaredhirsch:
Looks like our highlight stealing code does not work any more. FF 45.0.2, Mac.

Going to dig into why it's broken tomorrow.

jaredhirsch:
The duplicate checking code broke highlight management. With that change backed out, I'm able to consistently remove the highlight (with some exceptions, filed as separate sub-bugs).

Should have a PR out soon.

PR #144, opened 2016-04-18
118 remove initial highlight

jaredhirsch:
@chuckharmston R? if you're around, if not I'll land this tomorrow.

Bug #118 asks that we always steal the highlight when the popup is opened or new results are rendered. That's done here, with caveats: highlight stealing doesn't work if only the 'heuristic' top result is shown (#140); highlight stealing fails if the popup opens under the mouse pointer (#143). I'll tackle those issues in followup PRs.

Aside from making the highlighting uniform across recommendation shown and unshown cases, the main improvement here is better highlight stealing! Since we don't need to know if the recommendation is visible before stealing the highlight, it seems that our code wins the race more often. I also added three more nested `requestAnimationFrame` calls (now a total of 6). The final result is really nice: the blue highlight is much harder to summon, at least on my machine.

### DIFF
--- a/lib/ui/highlight-manager.js
+++ b/lib/ui/highlight-manager.js
@@ -43,6 +43,7 @@ HighlightManager.prototype = {
     this.resultsObserver.disconnect();
     this.detachMouseListeners();
 
+    this.events.unsubscribe('recommendation-created', this.initMutationObserver);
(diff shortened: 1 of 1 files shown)


jaredhirsch:
Works, looks reasonable, merging.

PR #146, opened 2016-04-20
Adjust XUL popup state, fixing popup in Nightly

### LINKED ISSUE #138: Awesomebar results broken in the latest nightly with "this.mInput is null"
I get this in the browser console:

```
TypeError: this.mInput is null autocomplete.xml:1140:11

get__matchCount() autocomplete.xml:1140
_invalidate() autocomplete.xml:1107
invalidate() autocomplete.xml:1097
Search.prototype.notifyResults() UnifiedComplete.js:1821
_addMatch() UnifiedComplete.js:1443
Search.prototype._onResultRow() UnifiedComplete.js:1360
bound () self-hosted
ConnectionData.prototype<._executeStatement/pending<.handleResult() Sqlite.jsm:763

```

I bet this is caused by the landing of [bug 1195054](https://bugzilla.mozilla.org/show_bug.cgi?id=1195054).

jaredhirsch:
Fixes #138.

Commit message below summarizes the situation reasonably well. @chuckharmston, R? if you're around, if not, I'll review, double-check, and land this myself tomorrow.

Commit message follows:

---

First, when the popup opens for the first time, its `mInput` will be
null, which breaks rendering of all the results. To work around this,
detect when the popup opens for the first time, via a MutationObserver
on the popup's 'hidden' attribute, then set the missing `mInput` state.

Second, each time the popup closes, its `mPopupOpen` state needs to be
reset to false. This is normally done by the `popuphiding` listener
defined in XBL. However, because our code defines its own `popuphiding`
listener, it seems that the XBL handler is not called. Work around this
by temporarily detaching our listener, firing a synthetic `popuphiding`
event on the XUL popup element, then reattaching the listener.

Note that I'm not convinced attaching a DOM Level 2 event listener
should ever have the effect of breaking an XBL (pseudo-DOM Level 0)
handler. Treat the explanation in the previous paragraph as a
hypothesis, not as a theorem :-P

Fixes #138.

### DIFF
--- a/lib/ui/popup.js
+++ b/lib/ui/popup.js
@@ -12,6 +12,7 @@ function Popup(opts) {
   this.win = opts.win;
   this.events = opts.events;
   this.mouseOverTimeout = null;
+  this.popupOpenObserver = null;
 
   // Public property used when we need to ignore a single mouseover event,
   // due to the mouse pointer being positioned over the popup when the results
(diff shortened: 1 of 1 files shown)


lib/ui/popup.js
```diff
@@ -20,12 +21,27 @@ function Popup(opts) {
...
 Popup.prototype = {
   init: function() {
     this.el = this.win.document.getElementById('PopupAutoCompleteRichResult');
     this.el.addEventListener('popuphiding', this.beforePopupHide);
+
+    // The first time the popup opens, a bit of popup state is not set
+    // correctly, causing no results to be shown, and causing the popup to fail
+    // to appear more than once (#138). To work around this bug, listen for the
+    // popup to be opened for the first time, and set the missing state.
+    //
```
chuckharmston, line 36: Perhaps a link to the relevant Gecko code this is emulating would be nice for future maintainers?

chuckharmston:
Played with it for 15 minutes and it all seemed to work well. There was a single `this.mInput is null` error in the console that I have been unable to reproduce, so...I think it's fine?

chuckharmston:
Nice detective work! Merge at will.

lib/ui/popup.js
jaredhirsch, line 36: Yeah, that's a good point. I'll try to figure something out. Unfortunately, the source of this bug is that the `richlistitem` definition in `autocomplete.xml` looks like it's been copied into `urlbarBindings.xml` and slightly changed. Because the copying over and the additional changes are mixed in the same commit (https://github.com/mozilla/gecko-dev/commit/ee27759c), you have to wade through ~1800 lines of changes to try to make sense of what's been altered. I've been staring at this code for nearly a year now, on and off, and I wasn't able to make much progress--hence falling back to this PR that merely fixes the surface symptoms :-\

I'll probably link to the gecko-dev commit and add a warning comment that it is a large, complex diff.

jaredhirsch:
Thanks for looking, @chuckharmston! Merging

PR #154, opened 2016-04-27
Detect mouse movement via mousemove, not mouseover

### LINKED ISSUE #143: Highlight stealing fails if the popup opens under the mouse

jaredhirsch:
@chuckharmston R? I've tried to thoroughly document the changes here. Please let me know if anything seems unclear.

Verifying this PR works: position the mouse pointer so that, when you type in the urlbar, the popup opens under the pointer. Without this PR applied, highlight stealing will break due to spurious mouseover events, causing the top result to take the blue highlight. With this PR applied, highlight stealing should work.

Original commit message follows:

---

Effective highlight stealing requires adjusting the highlight over many
(currently 6) successive turns for each DOM Mutation observed as results
stream into the popup. Because the stealing spans 6 \* 17ms ~= 100ms, we
need to detect when the mouse is moving and stop stealing the highlight
immediately. We used to do this by listening for mouseover events on the
popup, then canceling further highlight stealing until the popup
contents were updated.

When the popup opens under the mouse pointer, one or more spurious
mouseover events are fired. This makes highlight stealing buggy: we can
ignore a single mouseover event, but there's no easy way to know how
many mouseovers will be fired when the DOM is updated with new results.

In contrast, mousemove events are not spuriously fired when the popup is
opened under the mouse pointer. Mousemove events fire even more often
than mouseover events, so they are still a great tool to detect when the
user is actively mousing through the popup (the case where we want to
abort highlight stealing, even if results are still streaming in).
- Rename mouseover listener / variable names to mousemove.
- Remove the 'ignoreNextMouseOver' state on the popup; it's not needed,
  now that we are using mousemove.
- Update code docs.

Fixes #143.

### DIFF
--- a/lib/ui/popup.js
+++ b/lib/ui/popup.js
@@ -11,16 +11,11 @@ function Popup(opts) {
   */
   this.win = opts.win;
   this.events = opts.events;
-  this.mouseOverTimeout = null;
+  this.mouseMoveTimeout = null;
   this.popupOpenObserver = null;
 
-  // Public property used when we need to ignore a single mouseover event,
-  // due to the mouse pointer being positioned over the popup when the results
-  // list scrolls itself. See HighlightManager for more.
-  this.ignoreNextMouseOver = false;
-
   this.beforePopupHide = this.beforePopupHide.bind(this);
-  this.onResultsMouseOver = this.onResultsMouseOver.bind(this);
+  this.onResultsMouseMove = this.onResultsMouseMove.bind(this);
   this.onFirstPopupOpen = this.onFirstPopupOpen.bind(this);
 }
 
@@ -49,7 +44,7 @@ Popup.prototype = {
   destroy: function() {
     this.el.removeEventListener('popuphiding', this.beforePopupHide);
 
-    this.win.clearTimeout(this.mouseOverTimeout);
+    this.win.clearTimeout(this.mouseMoveTimeout);
 
     delete this.el;
     delete this.win;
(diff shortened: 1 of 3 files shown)


lib/ui/popup.js
```diff
@@ -82,20 +77,15 @@ Popup.prototype = {
...
+    // the mouse.
+    if (this.mouseMoveTimeout) {
       return;
     }
-    this.mouseOverTimeout = this.win.setTimeout(() => {
-      this.win.clearTimeout(this.mouseOverTimeout);
-      this.mouseOverTimeout = null;
+    this.mouseMoveTimeout = this.win.setTimeout(() => {
+      this.win.clearTimeout(this.mouseMoveTimeout);
+      this.mouseMoveTimeout = null;
```
chuckharmston, line 88: This is very clever, great idea.

lib/ui/popup.js
```diff
@@ -82,20 +77,15 @@ Popup.prototype = {
...
     }, 15);
```
chuckharmston, line 89: `1000/60` would be the technically correct value for this, but close enough `^.^`

chuckharmston:
Code looks great. Playing with it now...

chuckharmston:
Does this need rebased against master? I'm seeing a `this.mInput is null` exception on occasion.

chuckharmston:
I am able to reliably reproduce this condition with the patch applied, but not on master:

screen shot 2016-04-27 at 10 45 47 am

STR:
1. Initialize `window.universalSearch.popup.el` (i.e. get the popup to open and close once)
2. Type "f" in to the address bar.
3. Note that a recommendation is returning from the server.
4. Note that our inserted hbox is empty and is not collapsed.

chuckharmston:
Other than that, this looks and works great. If you can chase that down and it's a small fix, feel free to merge without review. :+1:

jaredhirsch:
Cool, thanks chuck!

jaredhirsch:
> Does this need rebased against master? I'm seeing a this.mInput is null exception on occasion.

I'm not sure why, but that error still appears sometimes. I'm going to find some time to discuss the `mInput is null` bug with the desktop search team, and hopefully better understand what's up.

jaredhirsch:
@chuckharmston I've tried, but I can't reproduce the error you were seeing. (I've actually seen that error in passing, but don't know how to reproduce it consistently.)

I'm going to merge this PR, but _please_ open a bug if you find reliable steps to reproduce the empty recommendation bug.

PR #160, opened 2016-04-29
Don't send cookies when making server requests

### LINKED ISSUE #148: Don't send cookies along with xhr
Security suggested we not send cookies when requesting data from the server.

Thanks to [a comment at the bottom of an old article](http://www.michael-noll.com/tutorials/cookie-monster-for-xmlhttprequest/), I discovered that Gecko provides a non-standard `LOAD_ANONYMOUS` flag (see [dxr](https://dxr.mozilla.org/mozilla-central/source/netwerk/base/nsIRequest.idl#205-210)) that we can use to prevent cookies being sent with our requests:

``` js
xhr.channel.loadFlags = Ci.nsIChannel.LOAD_ANONYMOUS;
```

jaredhirsch:
@chuckharmston Mind taking a look? Security folks asked that we not send cookies when making requests from the recommendation server. Took a little digging, but it turns out that Gecko provides some non-standard flags with xhr, one of which strips cookies from the request: https://dxr.mozilla.org/mozilla-central/source/netwerk/base/nsIRequest.idl#205-210

This PR adds those flags in the spots where we make requests. You can verify the cookies are not being sent in the Network tab of the Browser Toolbox.

Fixes #148.

### DIFF
--- a/lib/recommendation-server.js
+++ b/lib/recommendation-server.js
@@ -4,6 +4,8 @@
 
 const EXPORTED_SYMBOLS = ['RecommendationServer']; // eslint-disable-line no-unused-vars
 
+const Ci = Components.interfaces;
+
 function RecommendationServer(opts) {
   /*
   Class providing a connection to a recommendation server and a method for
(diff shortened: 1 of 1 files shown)


chuckharmston:
Nice! This is really neat.

PR #161, opened 2016-04-29
Increase recommendation hiding timeout

### LINKED ISSUE #122: Increase the timeout before hiding stale results.
We currently allow stale recommendations to be shown for up to 50ms, to allow time for new recommendations to come in, preventing an unsightly highlight flash and some UI jank. Under the current deployment—without HTTP/2, a CDN, or an upstream cache—this is not enough. Let's increase it to a reasonable number—100-150ms should be fine—until we have a little more speed.

giphy

When this is merged, file a bug to revert this change once that happens.

jaredhirsch:
@chuckharmston R? Should be a quick one

Fixes #122.


chuckharmston:
💯

PR #163, opened 2016-04-30
Load images using Fetch, to avoid sending cookies to third parties

### LINKED ISSUE #150: Load images using Fetch or xhr
By default, the browser sends cookies when it fetches an image via an Image tag. For privacy reasons, we don't want to do this.

Instead, let's use Fetch to download images, then stuff the data into an image tag on the client:

``` js
fetch(imgURL, { mode: 'no-cors' }).then(insertIntoDOM)
```

jaredhirsch:
@chuckharmston R?

Use Fetch to avoid leaking cookies when loading images. Code largely lifted from https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch.

Fixes #150.

### DIFF
--- a/lib/ui/recommendation-row.js
+++ b/lib/ui/recommendation-row.js
@@ -52,10 +52,23 @@ RecommendationRow.prototype = {
     let favicon = row._get('enhancements.favicon.url');
     let defaultFavicon = 'chrome://mozapps/skin/places/defaultFavicon@2x.png';
     let elem = createElement(row.doc, 'image', 'favicon');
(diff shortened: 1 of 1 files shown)


chuckharmston:
Can we abstract this to a utility? It seems to work great, but as complexity grows, I think this will be a capability that we would end up copy/pasting all over the place.

Input: URL, output: `<image>` tag, fetched.

jaredhirsch:
@chuckharmston How about a followup bug for the first post-release milestone?

PR #176, opened 2016-05-03
Displays only predefined result types (closes #165).

### LINKED ISSUE #165: Only show recommendations for predefined result types.
- Create a formal interface/base for a "result type" class, which should serve a similar purpose to the current `RecommendationRow` class.
- Create Wikipedia and TLD result type classes.
  - Could use [Wikipedia W](https://commons.wikimedia.org/wiki/File:Wikipedia-W-bold-in-square.svg) in the first column?
- Attach result type classes to an object in the app namespace.
- Add a method to the `RecommendationServer` class that negotiates for and returns the appropriate result type class or `null`.
- Refactor `RecommendationServer.search` to only publish the `recommendation` event if that method returns a truthy value.
- Refactor `RecommendationServer.search` to include the appropriate result type class in the event payload.
- Refactor `Recommendation.show` to use the result type class from the event payload when rendering.

Blocked on https://github.com/mozilla/universal-search-recommendation/issues/117

chuckharmston:
r? @6a68

Video of this in action (along with https://github.com/mozilla/universal-search-recommendation/pull/118): https://chchc.ch/He06nt


jaredhirsch:
@chuckharmston This approach is ok, though it's a bit unfortunate that the conditional logic is spread throughout the DOM rendering code. I [suggested](https://github.com/mozilla/universal-search/issues/165#issuecomment-216087894) centralizing the type-specific logic in the recommendation, and passing a view model-like config object into the recommendation row, in order to keep things a bit cleaner.

That said, we're in a hurry, and this works without excessive abstraction. I'll do a syntax nit-picking pass, make sure it's working, then turn it over to you to fix and merge.

chuckharmston:
In 03e02f2 I piggybacked a small UI fix that helps the jank problem immensely. [Before](https://chchc.ch/He06nt) / [After](https://chchc.ch/CSdYxa)

lib/ui/recommendation.js
```diff
@@ -141,15 +141,25 @@ Recommendation.prototype = {
...
       win: this.win
     }).render();
+
     if (this.el) {
-      if (this.el.firstChild) {
-        this.el.replaceChild(recommendation, this.el.firstChild);
+      if (recommendation) {
+        if (this.el.firstChild) {
+          if (urlOf(this.el.firstChild) !== urlOf(recommendation)) {
+            this.el.replaceChild(recommendation, this.el.firstChild);
```
jaredhirsch: We're already doing a similar check on the JS side: https://github.com/mozilla/universal-search/blob/master/lib/ui/recommendation.js#L130-L136

How about updating the existing duplicate check to just compare URLs?

chuckharmston: Hm, I didn't realize we were doing that. Makes sense, though. Fix incoming.

chrome/skin/style.css
```diff
@@ -154,6 +154,13 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
   background-image: linear-gradient(rgba(255,255,255,0.3), transparent);
   color: HighlightText;
 }
+#universal-search-recommendation.highlight [data-result-type="wikipedia"] #universal-search-recommendation-icon {
```
jaredhirsch: It looks like this data- attribute is only used for styling. Let's use a classname instead. If we need to get the result type, we can get it on the JS side (`recommendation.resultType`) without paying the high price of accessing the DOM.

lib/ui/recommendation-row.js
```diff
@@ -30,32 +35,52 @@ RecommendationRow.prototype = {
     recommendation.
     */
     let row = this;
+
+    if (!this.resultType) {
+      return null;
```
jaredhirsch: Rather than return a specific value like `return null`, you can generally just `return;` in JS. The default return value is `undefined`, which is falsy, and that's often enough to get the job done.

lib/ui/recommendation-row.js
```diff
@@ -30,32 +35,52 @@ RecommendationRow.prototype = {
...
+  getResultType: function(row) {
+  /*
+  Determines and returns a TYPES property indicating the result type.
+  */
+    if (row._get('enhancements.tld')) {
+      return TYPES.TLD;
+    } else if (row._get('enhancements.wikipedia')) {
+      return TYPES.WIKIPEDIA;
+    }
+    return null;
```
jaredhirsch: Ditto here, just `return;`

lib/ui/recommendation-row.js
```diff
@@ -30,32 +35,52 @@ RecommendationRow.prototype = {
...
   renderFavicon: function(row) {
     /*
     Creates and returns an <image> element representing the favicon. If none is
     available, or if the URL returned from the server errors, will use
     Firefox's default favicon.
     */
-    let favicon = row._get('enhancements.favicon.url');
     let defaultFavicon = 'chrome://mozapps/skin/places/defaultFavicon@2x.png';
-    let elem = createElement(row.doc, 'image', 'favicon');
+    let wikipedia = 'chrome://universalsearch-skin/content/wikipedia-dark.svg';
```
jaredhirsch: Should this be a constant declared up at the top of the file?

lib/ui/recommendation-row.js
```diff
@@ -30,32 +35,52 @@ RecommendationRow.prototype = {
...
+  },
+
   renderFavicon: function(row) {
     /*
     Creates and returns an <image> element representing the favicon. If none is
     available, or if the URL returned from the server errors, will use
     Firefox's default favicon.
     */
-    let favicon = row._get('enhancements.favicon.url');
     let defaultFavicon = 'chrome://mozapps/skin/places/defaultFavicon@2x.png';
```
jaredhirsch: May as well make this a constant too

lib/ui/recommendation-row.js
```diff
@@ -194,8 +223,11 @@ RecommendationRow.prototype = {
     recommendation's URL.
     */
     let elem = createElement(row.doc, 'description', 'url');
-    elem.textContent = row._unescape(row._get('result.url'));
-    elem.setAttribute('data-url', row._get('result.url'));
+    let url = row._unescape(row._get(row.resultType == TYPES.WIKIPEDIA ?
+                                     'enhancements.wikipedia.url' :
+                                     'result.url'));
```
jaredhirsch: I think it's best to stick with the more verbose style from lines 212-216, just for consistency

lib/ui/recommendation-row.js
jaredhirsch: actually, this return isn't needed at all. it'll return undefined by default.

lib/ui/recommendation-row.js
```diff
@@ -30,32 +35,52 @@ RecommendationRow.prototype = {
...
     Creates and returns an <image> element representing the favicon. If none is
     available, or if the URL returned from the server errors, will use
     Firefox's default favicon.
     */
-    let favicon = row._get('enhancements.favicon.url');
     let defaultFavicon = 'chrome://mozapps/skin/places/defaultFavicon@2x.png';
-    let elem = createElement(row.doc, 'image', 'favicon');
+    let wikipedia = 'chrome://universalsearch-skin/content/wikipedia-dark.svg';
+    let favicon = (row.resultType == TYPES.WIKIPEDIA ? wikipedia :
+      (row._get('enhancements.favicon.url') || defaultFavicon));
```
jaredhirsch: I think it'd be best to convert this to the more verbose, clearer style on lines 212-216.

jaredhirsch:
@chuckharmston Alright! Nice work. You're good to merge once these changes are made, provided the code works with the recommendation server branch locally :+1:

lib/ui/recommendation.js
```diff
@@ -142,21 +143,29 @@ Recommendation.prototype = {
...
-      this.data = data;
-      this.events.publish('recommendation-shown', this);
     }
   },
   hide: function() {
     if (this.el) {
       const container = this.win.document.getElementById(RECOMMENDATION_ID);
       this.timeout = this.win.setTimeout(function() {
-        this.el.collapsed = true;
+        //this.el.collapsed = true;
```
jaredhirsch, line 168: Oops

lib/ui/recommendation.js
```diff
@@ -129,7 +129,8 @@ Recommendation.prototype = {
 
     // Don't re-render the recommendation if the content hasn't changed since
     // the previous recommendation.
-    if (this.objectUtils.deepEqual(data, this.prev)) {
+    if (this.prev && this.el.firstChild && data.result.url === this.prev.result.url) {
+      this.win.console.log('Not re-rendering', this.prev);
```
jaredhirsch, line 133: Was this console.log intended to be left here?

chuckharmston, line 133: Nope, sorry, I'm rushing pretty hard to get this out today.

PR #177, opened 2016-05-03
Add Test Pilot URL to description

### LINKED ISSUE #174: Add reference to testpilot.firefox.com in addon meta
We have to add something to the `about:addons` page that 
1. Links back to https://testpilot.firefox.com.
2. Identifies this a Test Pilot experiment.

Exact specifications provided by @nchapman, @johngruen

jaredhirsch:
@chuckharmston quick R?

Fixes #174.

Screenshot in my about:addons with the line added:

screen shot 2016-05-03 at 12 47 44 pm

### DIFF
--- a/install.rdf
+++ b/install.rdf
@@ -7,7 +7,7 @@
     <em:unpack>false</em:unpack>
     <em:version>1.0.2</em:version>
     <em:name>Universal Search</em:name>
-    <em:description>Universal Search puts the best sites from around the web in your Awesome Bar. By typing just a few characters, you can easily navigate to the best the web has to offer. You’ll also find recommendations for popular people and articles from Wikipedia so you can quickly get answers to your burning questions.</em:description>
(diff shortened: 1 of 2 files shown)


jaredhirsch:
@chuckharmston Actually, wait just a minute, I'll add the 1.0.3 release to the branch as well

chuckharmston:
What's the point of having that URL there if it's not clickable? Can we just change the homepage instead?

chuckharmston:
Oops, just saw the comments in the bug. Kind of crappy, still. @nchapman can we just change `em:homepageURL` instead and make reference to Test Pilot?

jaredhirsch:
> What's the point of having that URL there if it's not clickable? Can we just change the homepage instead?

@chuckharmston See product request here https://github.com/mozilla/universal-search/issues/174#issuecomment-216641543

jaredhirsch:
@chuckharmston I'm fine to put the text in both places, whatever

chuckharmston:
Let's ship it so we don't have to block on product. We'll be doing another release soon for #165 anyway, we can fix then.

Issue #181, opened 2016-05-03
Sometimes get the wrong results when I backspace


ghost:
This isn't 100% reproducible, but I tried it on a brand new profile (see recording below) and got it on the second try.

STR:
1) Type 'fac' to get facebook to prefill
2) Backspace over everything and type 'h'
3) See Facebook recommended

![out-12](https://cloud.githubusercontent.com/assets/16946141/15001411/1ac2f088-114c-11e6-9745-f6aec3f450f7.gif)

(edit: lol @ gif quality.  What are you kids using for web animations these days? webm?)

jaredhirsch:
Marking as public launch blocking just so we all notice it this week.

chuckharmston:
I have not been able to reproduce this at all.

@wresuolc Can I ship you a fancy special build with additional logging output?

ghost:
sure

chuckharmston:
@wresuolc added a debugging build here: https://gist.github.com/chuckharmston/e4e2c51ac031eb80f6194735a1814683#debugging-build

I'm particularly interested in any logs surrounding a failed recommendation. It's pretty noisy; filtering for `universal-search` will help you make sense of it:

![screen shot 2016-05-04 at 11 26 27 am](https://cloud.githubusercontent.com/assets/23885/15022894/10559192-11eb-11e6-96c3-d7ebe26b5f3d.png)

ghost:
I installed the build in a new profile.  On a fresh restart the awesomebar works once and I see a suggestion for facebook, and then when I backspace I see this in the console (among lots of other things):

```
JavaScript error: chrome://browser/content/urlbarBindings.xml, line 1446: TypeError: this.searchSuggestionsNotification is null
JavaScript error: chrome://universalsearch-ui/content/recommendation-row.js, line 259: TypeError: can't convert null to object
JavaScript error: chrome://browser/content/urlbarBindings.xml, line 1446: TypeError: this.searchSuggestionsNotification is null
JavaScript error: chrome://browser/content/urlbarBindings.xml, line 1446: TypeError: this.searchSuggestionsNotification is null
```

and the last line just repeats forever and the awesomebar stops working completely.

chuckharmston:
> JavaScript error: chrome://universalsearch-ui/content/recommendation-row.js, line 259: TypeError: can't convert null to object

This is the one that I needed. Thanks! Fix will be incoming.

PR #183, opened 2016-05-04
Metrics fo realz

jaredhirsch:
**Update: no longer work in progress, this bad boy is ready to merge**

---

WIP commit, still needs a little cleanup. @chuckharmston mind taking a look?

Opening as WIP to make sure we're gathering enough data for the first pass. Comments welcome.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -7,6 +7,7 @@ A summary of the metrics the Universal Search add-on and recommendation server w
 - **URL bar** — the unified address + search bar at the top of the browser.
 - **Popup** — the popup that appears when you begin typing into the URL bar.
 - **Search** — the full lifecycle, beginning when the user types a key into the url bar, and ending when the user either chooses an item from or otherwise dismisses the popup.
+- **Recommendation** - the recommendation UI inserted into the popup by this add-on.
 
 
 ## Data collection
@@ -16,36 +17,73 @@ Metrics gathered from the Universal Search add-on will be reported to the Test P
 
 ### On the client
 
-Universal Search will record and report each of the following events:
+#### The popup lifecycle
 
-- User selects a recommendation.
-	- Time from search start to search completion.
-	- Length of query when selected.
-	- List of enhancers applied (e.g. `Object.keys(recommendation.enhancements)`.
-	- Value of the recommendation's `data-type` attribute. _(NB: this likely won't matter in early experiments. This is intended to allow explicit monitoring of qualitative types of results, e.g. "movie cards".)_
-- User enters a URL.
-	- Time from search start to search completion.
-- Use directly searches using their default search engine.
-	- Time from search start to search completion.
-	- Length of query.
-- User selects a suggested search term.
-	- Time from search start to search completion.
-	- Length of query.
-	- Levenstein distance between query and completed search term.
-- User selects a bookmark.
-	- Time from search start to search completion.
-	- Length of query.
-- Use selects an item from history.
-	- Time from search start to search completion.
-	- Length of query.
-- The user terminates a search.
+- The popup usually opens when the user types in the urlbar. It also opens in some less common circumstances, like clicking the history drop marker (the small down caret to the right of the urlbar).
+- The popup closes when the user navigates or dismisses the popup without navigating.
+  - The popup can be dismissed without navigating by hitting the 'Escape', left arrow, or right arrow keys.
+  - The user can navigate by clicking a result in the popup, or by using the arrow or Tab keys to highlight a result in the popup, then hitting Enter.
+  - The user can also navigate by typing in the urlbar and hitting Enter, without ever interacting with the popup.
 
-This data will be used in three ways:
+#### Measuring the popup lifecycle
 
-1. To gain a deeper understanding of the value of our recommendations to users.
-2. To gain a deeper understanding as to how users engage with the URL bar, search popup, and our recommendations.
-3. To segment users of other metrics.
+Each time the popup is closed, we record some facts about Firefox's state:
 
+- `didNavigate`: was the popup closed because the user navigated, or because the popup was dismissed without navigating (for example, using the Escape key)?
+- `interactionType`: if the user did navigate, was the selected item clicked or keyed?
+- `recommendationShown`: was a recommendation visible when the popup was closed?
+- `recommendationType`: if a recommendation was shown, was it a wikipedia or a top-level domain recommendation?
+  - These types will grow over time, as we add new kinds of recommendations.
+- `recommendationSelected`: if the user navigated, and if the recommendation was visible, did the user navigate to the recommendation URL?
+- `selectedIndex`: if the user navigated to an item in the popup, which item was selected?
+  - For instance, if the third non-recommendation result was clicked, `selectedIndex` would have a value of 2.
+  - If the recommendation was selected, the `selectedIndex` value will be -1.
+
+Here's a rough schema:
+
+```js
+{
+  didNavigate: true or false
+  interactionType: 'click' or 'key'
+  recommendationShown: true or false
+  recommendationSelected: true or false
+	recommendationType: 'tld' or 'wikipedia'
+  selectedIndex: an integer between -1 and 30 inclusive
+}
+```
+
+Here's an example of a complete telemetry ping:
+
+```js
+{
+	"test": "universal-search@mozilla.com",  // The em:id field from the add-on
+	"agent": "User Agent String",
+	"payload": {
+		"didNavigate": true,
+		"interactionType": "click",
+		"recommendationShown": true,
+		"recommendationType": "tld",
+		"recommendationSelected": true,
+		"selectedIndex": -1
+	}
+}
+```
+
+#### Data Analysis
+
+We will use this data to calculate click-through rates (CTR) for the popup as shown below.
+
+- CTR of the popup overall: number of events with `didNavigate = true` / total event count
+- CTR when recommendations are shown: (number of events with `didNavigate = true` and `recommendationShown = true`) / (total events with `recommendationShown = true`)
+- CTR when recommendations are not shown: (number of events with `didNavigate = true` and `recommendationShown = false`) / (total events with `recommendationShown = false`)
+- CTR when wikipedia recommendations are shown: (number of events with `didNavigate = true` and `recommendationType = wikipedia`) / (total events with `recommendationType = wikipedia`)
+- CTR when TLD recommendations are shown: (number of events with `didNavigate = true` and `recommendationType = tld`) / (total events with `recommendationType = tld`)
+- When users navigate via the popup, how often do they use the mouse: (number of events with `didNavigate = true` and `interactionType = click`) / (total events with `didNavigate = true`)
+- When users navigate via the popup, how often do they use the keyboard: (number of events with `didNavigate = true` and `interactionType = key`) / (total events with `didNavigate = true`)
+- CTR for the nth result down the list: (number of events with `didNavigate = true` and `selectedIndex == n`) / (total events with `didNavigate = true`)
+  - We'll use this formula to generate a histogram for the various values of `selectedIndex`.
+
+By using CTR as a proxy for helpfulness, we can use these CTRs to measure whether recommendations are helpful overall, and which types are most or least helpful.
 
 ### On the server
 
--- a/lib/metrics.js
+++ b/lib/metrics.js
@@ -1,0 +1,128 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const EXPORTED_SYMBOLS = ['Metrics'];
+
+function Metrics(opts) {
+  /*
+  The metrics class measures the state of the popup each time it closes.
+  Data is sent to the Test Pilot add-on, which forwards it to the Telemetry
+  server.
+  */
+  this.events = opts.events;
+  this.interaction = null;
+  this.observerService = opts.observerService;
+  this.win = opts.win;
+  this.recommendation = opts.recommendation;
+  this.popup = opts.popup;
+  this.checked = false;
+
+  this.interactionStart = this.interactionStart.bind(this);
+  this.interactionEnd = this.interactionEnd.bind(this);
+  this.onKeyNavigate = this.onKeyNavigate.bind(this);
+  this.onClickNavigate = this.onClickNavigate.bind(this);
+}
+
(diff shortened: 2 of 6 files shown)


jaredhirsch:
One outstanding TODO: this needs to be rebased against master, and we need to include the recommendation type (wikipedia vs tld) in the packet that's sent down.

lib/metrics.js
```diff
@@ -0,0 +1,129 @@
...
+
+      // What type of recommendation was it?
+      recommendationType: null,
+
+      // Was the recommendation the selected item?
+      recommendationSelected: null,
+
+      // What was the index of the selected item?
+      // If the recommendation was selected, this will be -1.
+      selectedIndex: null
```
jaredhirsch, line 72: @nchapman Does this look like it answers the basic product questions? This packet is constructed each time the popup opens, and is fired each time the popup closes, regardless of whether the user navigated or interacted with the popup while it was open. One missing bit of data that I'll add is the wikipedia vs TLD type.

This data should be enough for us to calculate the click-through rate of the popup with and without the recommendation visible, and drill down into click-through rates of the different types of recommendations.

We should also be able to figure out how often users interact with the popup, since it always pops open when you enter stuff in the urlbar.

jaredhirsch:
@lmorchard @clouserw This metrics patch is coming together, comments welcome if you have spare cycles. The way I'm doing it is fast and simple, but will require writing some code to calculate KPI percentages on the server. Since we'll have some flexibility in data visualization, it seems like this should work fine.

lib/metrics.js
```diff
@@ -0,0 +1,129 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const EXPORTED_SYMBOLS = ['Metrics'];
+
+function Metrics(opts) {
+  /*
+  The metrics class is used to determine whether the recommendation is used
+  when the user navigates with the popup open. Data is sent to the Test Pilot
```
jaredhirsch: This is not quite accurate. We fire a signal every time the popup closes, whether or not the user navigated or hit `Escape` to dismiss the popup.

lib/metrics.js
```diff
@@ -0,0 +1,129 @@
...
+
+  this.interactionStart = this.interactionStart.bind(this);
+  this.interactionEnd = this.interactionEnd.bind(this);
+  this.onKeyNavigate = this.onKeyNavigate.bind(this);
+  this.onClickNavigate = this.onClickNavigate.bind(this);
+}
+
+Metrics.prototype = {
+  init: function() {
+    this.win.console.log('metrics.init');
```
jaredhirsch: All of the console.log() calls will be removed when this graduates from WIP to real thing

lib/metrics.js
```diff
@@ -0,0 +1,129 @@
...
+    this.events.unsubscribe('before-popup-hide', this.interactionEnd);
+  },
+  interactionStart: function() {
+    this.win.console.log('metrics.interactionStart');
+    // Initialize default values for the case where the popup closes without a
+    // navigation.
+    this.checked = false;
+    this.interaction = {
+
+      // Did the user navigate?
```
6a68: Better: "Did the popup close because the user navigated to a URL?"

lib/metrics.js
```diff
@@ -0,0 +1,129 @@
...
+    this.interaction = {
+      // If interactionType is defined, then we did navigate.
+      didNavigate: !!interactionType,
+      interactionType: interactionType || null,
+      recommendationShown: !this.recommendation.el.collapsed,
+      recommendationSelected: this.recommendation.isHighlighted,
+      recommendationType: this.recommendation.resultType,
+      // If no selectedIndex was passed in, we get the value from the popup.
+      // Note that `isNaN(undefined) => true`.
+      selectedIndex: isNaN(selectedIndex) ? this.popup.el.selectedIndex : selectedIndex
```
jaredhirsch: Hm. It's possible that the user might hit a letter and the enter key faster than the popup can open for the first time. In that case, `this.popup.el` might not exist yet, so that this would throw.

lib/metrics.js
```diff
@@ -0,0 +1,129 @@
...
+    this.win.setTimeout(() => {
+      this.win.console.log('metrics.interactionEnd');
+      // If we haven't checked state, check it now, so we have data to report.
+      if (!this.checked) {
+        this.checkState();
+      }
+      const data = {
+        id: 'universal-search@mozilla.com',
+        data: JSON.stringify(this.interaction)
+      };
```
jaredhirsch, line 117: oops, this needs to be { subject, data } https://github.com/mozilla/testpilot/blob/master/addon/lib/metrics.js#L111

lib/ui/recommendation.js
```diff
@@ -167,6 +167,12 @@ Recommendation.prototype = {
       return;
     }
 
+    // For metrics, record whether the event was a click or key.
+    const eventData = {
+      click: !!evt.button,
+      type: this.resultType
+    };
+
```
jaredhirsch: The changes in this file are stale (from an earlier attempt), ignore. I'll remove.

chuckharmston:
Metrics spec, for reference: https://github.com/mozilla/universal-search/blob/master/docs/metrics.md

lib/metrics.js
```diff
@@ -0,0 +1,129 @@
...
+  this.checked = false;
+
+  this.interactionStart = this.interactionStart.bind(this);
+  this.interactionEnd = this.interactionEnd.bind(this);
+  this.onKeyNavigate = this.onKeyNavigate.bind(this);
+  this.onClickNavigate = this.onClickNavigate.bind(this);
+}
+
+Metrics.prototype = {
+  init: function() {
```
chuckharmston, line 28: This ended up being much cleaner and less intrusive than I thought it would thanks to how you've architected this. Really nicely done.

lib/metrics.js
```diff
@@ -0,0 +1,129 @@
...
+      didNavigate: false,
+
+      // Was the selected item clicked or keyed or neither?
+      interactionType: null,
+
+      // Was the recommendation shown?
+      recommendationShown: false,
+
+      // What type of recommendation was it?
+      recommendationType: null,
```
chuckharmston, line 65: Notably, we also want a list of all the recommendation's enhancements `Object.keys(recommendation.enhancements)` if a recommendation is selected.

chuckharmston:
Do we have the capacity to determine the type of non-recommendation selections? e.g. bookmark vs history vs search vs suggested search term? That was a pretty [key part](https://github.com/mozilla/universal-search/blob/master/docs/metrics.md#result-types) of the initial idea, helping us understand how users interact with the awesome bar, in addition to our recommendation.

chuckharmston:
We also want to collect information about [which data is shown](https://github.com/mozilla/universal-search/blob/master/docs/metrics.md#rich-data). With the recent change to distinct result types, we may want to tweak what exactly this reports.

chuckharmston:
This is brilliant. Thank you!

lib/metrics.js
chuckharmston: We should probably consider a unified logging approach, though, to help facilitate error reporting. something like:

```
Logger.prototype = {
  _prefix: function() {
    return `[${this.classPrefix}.${this.instancePrefix}]`;
  },

  log: function() {
    [].splice.apply(arguments, [0, 0, this._prefix()]);
    this.win.console.log.apply(this.win.console, arguments);
  },

  warn: function() {
    [].splice.apply(arguments, [0, 0, this._prefix()]);
    this.win.console.warn.apply(this.win.console, arguments);
  },

  error: function() {
    [].splice.apply(arguments, [0, 0, this._prefix()]);
    this.win.console.error.apply(this.win.console, arguments);
  }
};


// Then in the Metrics class...
this.logger = new this.Logger('metrics', this.win);

// Then in Metrics.init, to log the Metrics instance:
this.logger.log('init', this);

// Console sez:
'[universal-search.metrics]' 'init' Object { ... }
```

jaredhirsch: Sending logger output to Telemetry sounds like a good idea, could help us figure out how frequently users hit errors. Maybe we can file a followup bug?

lib/metrics.js
jaredhirsch, line 28: :beers: Yeah, unobtrusive measurement is one of the nicest things about extremely late binding via pubsub (unit testing is another). If we had modules calling each other directly, or used promise-based APIs, it'd be a lot messier.

lib/metrics.js
chuckharmston: https://github.com/mozilla/universal-search/issues/184 :)

I'm working on this now so I can add some useful logging for @clouserw for #181.

lib/metrics.js
jaredhirsch, line 65: How about we file a followup bug for adding enhancements after public launch?

Based on the [examples](https://gist.github.com/chuckharmston/861a1442e7dde98aefe333a2347e0361) we discussed the other day, it looks like every TLD entry will have wikipedia, tld, favicon enhancements, while wikipedia entries will have wikipedia, favicon enhancements. This variation is captured by the recommendationType.

chuckharmston, line 65: That's not new scope, it's covered here: https://github.com/mozilla/universal-search/blob/master/docs/metrics.md#on-the-client

> List of enhancers applied (e.g. `Object.keys(recommendation.enhancements)`.

Notably, not every TLD entry will have a wikipedia enhancement—only if that search term has a Wikipedia page in the first set of results.

jaredhirsch:
Thanks for the feedback, @chuckharmston! There's definitely room to improve these metrics quite a bit, and I think we should file bugs for the first sprint post-public launch. This is intended to be a (very minimal) minimal viable metrics patch.

The data gathered by this patch lets us answer the fundamental questions product wants to answer:

**What is the click-through rate (CTR) of the popup overall?**
Given by: (events with `didNavigate = true`) / (total event count)

**What is the CTR when recommendations are shown?**
(events with `didNavigate = true` and `recommendationShown = true`) / (total events with `recommendationShown = true`)

**CTR when recommendations are not shown?**
Same as previous, with `recommendationShown = false`

**CTR when wikipedia recommendations are shown?**
(`didNavigate = true` and `recommendationType = wikipedia`) / (total events with `recommendationType = wikipedia`)

**CTR when TLD recommendations are shown?**
Same as previous, just swap out the `recommendationType`

**When users navigate via the popup, how often do they use the mouse?**
(`didNavigate = true` and `interactionType = click`) / (total events with `didNavigate = true`)

**When users navigate via the popup, how often do they use the keyboard?**
Same as previous, just swap out the `interactionType`.

**CTR for the 3rd result down the list?**
(`didNavigate = true` and `selectedIndex == 2`) / (total events with `didNavigate = true`)

I definitely want to instrument this more thoroughly, but I think this should be good enough for a first pass. cc @nchapman

_Edited to add CTR without recommendations shown and measure the selected index_

lib/metrics.js
nchapman, line 72: The two [KPIs identified](https://github.com/mozilla/universal-search/blob/master/docs/metrics.md#key-performance-indicators) are:
- the click through rate of the recommendations
- retention (how do people use recommendations over time)

and I think this let's us answer both those questions, so lgtm!

nchapman:
@6a68 this looks good! I commented, inline above, that as long as we can measure our 2 KPIs (CTR and retention) then we're good to go, and that very much appears to be true. We can add the secondary metrics in another pass.

jaredhirsch:
Great, I'll get this PR cleaned up and remove the "WIP" bit from the title, and we can land it today.

jaredhirsch:
@chuckharmston I guess this is not quite ready to land. I'm trying to figure out how to verify if this is working or not, because it's not clear exactly what syntax will work with the `notifyObservers` call. Going to try adding some logging to the test pilot add-on to verify.

docs/metrics.md
```diff
@@ -16,36 +17,73 @@ Metrics gathered from the Universal Search add-on will be reported to the Test P
...
-This data will be used in three ways:
-
-1. To gain a deeper understanding of the value of our recommendations to users.
-2. To gain a deeper understanding as to how users engage with the URL bar, search popup, and our recommendations.
-3. To segment users of other metrics.
+#### The popup lifecycle
 
+- The popup usually opens when the user types in the urlbar. It also opens in some less common circumstances, like clicking the history drop marker (the small down caret to the right of the urlbar).
+- The popup closes when the user navigates or dismisses the popup without navigating.
+  - The popup is dismissed by hitting the 'Escape', left arrow, or right arrow keys.
```
chuckharmston: ~~Also `enter`~~

Maybe `can be dismissed without navigating by`?

docs/metrics.md
```diff
@@ -16,36 +17,73 @@ Metrics gathered from the Universal Search add-on will be reported to the Test P
...
+```
+
+Here's an example of a complete telemetry ping:
+
+```js
+{
+	"test": "universal-search@mozilla.com",  // The em:id field from the add-on
+	"agent": "User Agent String",
+	"payload": {
+		“didNavigate”: true,
```
chuckharmston: Curly quotes on the key here.

docs/metrics.md
```diff
@@ -16,36 +17,73 @@ Metrics gathered from the Universal Search add-on will be reported to the Test P
...
+		"recommendationType": "tld",
+		"recommendationSelected": true,
+		"selectedIndex": -1
+	}
+}
+```
+
+#### Data Analysis
+
+We will use this data to calculate click-through rates for the popup as shown below.
```
chuckharmston: Nit: add `(CTR)` after `click-through rate`

lib/metrics.js
chuckharmston, line 65: `Notably` has apparently become a crutch word of mine.

chuckharmston:
Some documentation nits, but this looks great. 

> Going to try adding some logging to the test pilot add-on to verify.

If you do that and it seems to work, :shipit:!

docs/metrics.md
jaredhirsch: :+1:

docs/metrics.md
jaredhirsch: I have no idea how I did that. Trying again.

docs/metrics.md
jaredhirsch: :+1:

chuckharmston:
> I'm not sure under what circumstances this happens, but I've seen the error go by in the browser console.

I think this is just a timing error; because of the nature of XBL, it doesn't get inserted until the popup until the popup is instantiated for the first time, so any event-driven code that relies on it has to account for the fact that it may not quite be there. It's pretty easy to workaround, since it's an acceptable experience to just not show a recommendation.

PR #186, opened 2016-05-04
Uses correct add-on ID in update.rdf (closes #168).

### LINKED ISSUE #168: Add-on updates are not working
Debugging the auto-update process can be fiddly; `update.rdf` has to be served with the right mime type.

Let's make sure the update process works before public launch.

chuckharmston:
r? @6a68 

I'll make a release when this merges to make sure that additional internal users from the email blast today get updates.


jaredhirsch:
R+ :+1: :tada:

PR #188, opened 2016-05-05
Hardens RecommendationRow._get against null values (closes #181).

### LINKED ISSUE #181: Sometimes get the wrong results when I backspace

chuckharmston:
And a few drive-by semicolons.

r? @6a68

### DIFF
--- a/lib/ui/recommendation-row.js
+++ b/lib/ui/recommendation-row.js
@@ -256,8 +256,10 @@ RecommendationRow.prototype = {
     const val = key.split('.').reduce(function(o, x) {
       return (typeof o == 'undefined' || o === null) ? o : o[x];
     }, this._data);
-    const isEmptyObject = (typeof val == 'object' && !Object.keys(val).length);
-    return (typeof val == 'undefined' || !val || isEmptyObject) ? null : val;
+    let isEmptyObject = val => {
+      return val && typeof val == 'object' && !Object.keys(val).length;
+    };
+    return (typeof val == 'undefined' || !val || isEmptyObject(val)) ? null : val;
   },
 
   _firstLetterInBaseDomain: function(row) {
@@ -286,7 +288,7 @@ RecommendationRow.prototype = {
       return {
         'background': `rgb(${r}, ${g}, ${b})`,
         'foreground': brightness > 0.75 ? 'black' : 'white'
-      }
+      };
     }
     return;
   },
@@ -301,12 +303,12 @@ RecommendationRow.prototype = {
       return {
         'width': keyImage.width / keyImage.height * IMAGE_DIMENSION,
         'height': IMAGE_DIMENSION
-      }
+      };
     }
     return {
       'width': IMAGE_DIMENSION,
       'height': keyImage.height / keyImage.width * IMAGE_DIMENSION
-    }
+    };
   },
 
   _getContentMaxWidth: function(row) {
@@ -360,4 +362,4 @@ function createElement(doc, elementName, id) {
   let elem = doc.createElementNS(ns, elementName);
   elem.setAttribute('id', `universal-search-recommendation-${id}`);
   return elem;
-};
+}


lib/ui/recommendation-row.js
```diff
@@ -256,8 +256,10 @@ RecommendationRow.prototype = {
     const val = key.split('.').reduce(function(o, x) {
       return (typeof o == 'undefined' || o === null) ? o : o[x];
     }, this._data);
-    const isEmptyObject = (typeof val == 'object' && !Object.keys(val).length);
-    return (typeof val == 'undefined' || !val || isEmptyObject) ? null : val;
+    let isEmptyObject = val => {
+      return val && typeof val == 'object' && !Object.keys(val).length;
+    };
+    return (typeof val == 'undefined' || !val || isEmptyObject(val)) ? null : val;
```
jaredhirsch, line 262: The `typeof val !== 'undefined'` check is only needed if the variable `val` might be uninitialized, but since it's defined as a const above, you don't need it.

I think you can get away with just doing this:

``` js
return !!val && !isEmptyObject(val) && val || null;
```

lib/ui/recommendation-row.js
```diff
@@ -256,8 +256,10 @@ RecommendationRow.prototype = {
     const val = key.split('.').reduce(function(o, x) {
       return (typeof o == 'undefined' || o === null) ? o : o[x];
     }, this._data);
-    const isEmptyObject = (typeof val == 'object' && !Object.keys(val).length);
-    return (typeof val == 'undefined' || !val || isEmptyObject) ? null : val;
+    let isEmptyObject = val => {
+      return val && typeof val == 'object' && !Object.keys(val).length;
```
jaredhirsch, line 260: you can skip the truthy check here, since you do it on line 262 before this function is called: `return typeof val == 'object' && !Object.keys(val).length`

lib/ui/recommendation-row.js
```diff
@@ -256,8 +256,10 @@ RecommendationRow.prototype = {
     const val = key.split('.').reduce(function(o, x) {
       return (typeof o == 'undefined' || o === null) ? o : o[x];
     }, this._data);
```
jaredhirsch, line 258: Optional suggestion: how about this alternative for the first part? It handles all possible edge cases (like trying to access object properties on an array). It's slower (using try-catch sends you into interpreted vs compiled mode), but we're calling this in the midst of dealing with the DOM anyway, so speed's not a primary concern.

``` js
const val = key.split('.').reduce((o,x) => {
  let result;
  try {
    result = o[x];
  } catch (ex) {}
  return result;
}, this._data);
```

In the spirit of minimizing changes just before launch, I'm fine with ignoring this suggestion, but wanted to put it out there.

jaredhirsch:
@chuckharmston You know, it probably makes sense for us to rely on utils like lodash, rather than rolling our own without the benefit of unit tests. I've been resisting this idea out of laziness, but it makes more sense for us to spend an hour building a feature, rather than rewriting open source utils. There are definitely other spots in this addon that could benefit from replacement with open source tools.

Maybe I'll file a bug to replace this with lodash _.get after public launch, thoughts?

jaredhirsch:
@chuckharmston R+ from me once comments are addressed

lib/ui/recommendation-row.js
chuckharmston, line 260: I originally thought so too, but somehow `val` was `null` in #181. This is the only substantial change in PR; I just restructured to a quick function to make it cleaner.

jaredhirsch, line 260: Ah. If you look at the old line 259, you were doing the Object.keys(val) call without checking that val was truthy. `typeof null` is object, so that check didn't guard against null values.

chuckharmston, line 260: Right. That covers the two substantive comments here, so I'm going to merge this. Followed up with https://github.com/mozilla/universal-search/issues/199, which we should do very shortly after launch.

PR #190, opened 2016-05-05
Version 1.0.6

chuckharmston:
Getting this out at @wresuolc's request before internal marketing goes out today to make sure the metrics start flowing.

r? @6a68


chuckharmston:
Thanks Jared! 👍

PR #210, opened 2016-05-11
Correctly colors titles in Linux (closes #182).

### LINKED ISSUE #182: Wrong title color
Chuck asked for an issue about our title color on Linux.  It should be white?  Filing so we don't lose track -- we can triage later. :)

chuckharmston:
r? @6a68


jaredhirsch:
LGTM but I don't have a linux testing environment right now. Let's merge and see if @wresuolc notices a fix next time we do a release

Issue #200, opened 2016-05-10
Show special cards for searches recognized as movies.


chuckharmston:
Blocked on https://github.com/mozilla/universal-search-recommendation/issues/134

To be defined:
- [x] To what priority do we give movie hits relative to wikipedia or TLDs? (Probably: higher than each. @nchapman to confirm) 
- [x] What does it look like? (@bryanbell to come in with a design. Assigning to him)

chuckharmston:
TLD > Movie > Wikipedia

jaredhirsch:
Copying over @bryanbell's mock from https://github.com/mozilla/universal-search-recommendation/issues/134:

---

![rec - movie - captain](https://cloud.githubusercontent.com/assets/228979/15308052/038879b4-1b8e-11e6-9d02-28edc573b272.png)

jaredhirsch:
Here's where we'll have to leave it for a few weeks:

<img width="693" alt="screen shot 2016-05-27 at 5 06 05 pm" src="https://cloud.githubusercontent.com/assets/96396/15623759/5ec55a84-242d-11e6-9ccd-54f46043a023.png">

Highlighted:

<img width="696" alt="screen shot 2016-05-27 at 5 06 20 pm" src="https://cloud.githubusercontent.com/assets/96396/15623758/5ec365bc-242d-11e6-9172-573cef346d20.png">

The card looks ready to go, but I haven't had enough time to do a final cleanup pass, particularly on the CSS--we should probably only use XUL's box model, as mixing in HTML5 flexbox CSS properties is a recipe for madness. Blake suggested only using attributes for the weirdo XUL box model `pack` and `align` properties (these are currently also used via their CSS equivalents, `-moz-box-pack` and `-moz-box-align`). Code is stashed in the wip-movie-card branch on my fork for now. Also, our existing code increases the minimum width of the popup, but I think we should stop doing that, because the built-in popup does not grow its width to fit its content, and the width can jump around on a narrow window.

Here are some screenshots of native behavior and our recommendation UI on release channel (46.0.1):

1) Note that the Firefox native popup behavior is to never stretch past the boundaries of the urlbar. On dev edition/nightly it's similar, except the bounds are the window.

<img width="334" alt="screen shot 2016-05-27 at 4 45 41 pm" src="https://cloud.githubusercontent.com/assets/96396/15623662/b3891918-242b-11e6-93eb-68f0f72a7960.png">

[...]

PR #211, opened 2016-05-12
Refactor view layer to use distinct result types (closes #178).

### LINKED ISSUE #178: Refactor view layer to be more friendly to distinct result types.
Our view layer is pretty messy right now, and that will just get worse as we expand into more distinct result types in future sprints. Let's refactor to something class-based and more extensible after our public launch.

The groundwork for this was laid in #166, which was shelved.

chuckharmston:
It's baaaaaack. It'd be great to have this in place before diving into #200.

r? @6a68 

TODO:
- [ ] Write docstrings for the result classes.


jaredhirsch:
:eyes: looking now

lib/recommendation-server.js
```diff
@@ -82,17 +89,36 @@ RecommendationServer.prototype = {
...
   },
+
+  getResultClass: function(data) {
+    let types = Object.keys(data.enhancements);
+    if (types.indexOf('tld') > -1) {
+      return this.resultClasses.tld;
+    } else if (types.indexOf('wikipedia') > -1) {
+      return this.resultClasses.wikipedia;
+    }
+    return null;
```
jaredhirsch: You don't need to use `Object.keys` here. Just check for tld like this: `if ('tld' in data.enhancements)` and ditto for looking for the wikipedia key in the enhancements object.

There's no need to return null; undefined is returned by default, which works for falsy checks.

chuckharmston: I try to avoid `'str' in {}`, since it can fail wildly if the prototype is ever modified. Probably doesn't matter here, but `¯\_(ツ)_/¯`

lib/recommendation-server.js
```diff
@@ -82,17 +89,36 @@ RecommendationServer.prototype = {
...
+  },
+
   search: function(payload) {
+    let server = this;
```
jaredhirsch, line 93: In this repo, I've been systematically binding the value of `this` using two steps: 
- `bind()` the `this` value inside the constructor, if the function is called externally without the object reference (like event handler callbacks, setTimeout callbacks, or xhr success callbacks)
- use arrow fns for inner functions that can be anonymous (callbacks that never need to be detached by reference later)

It looks like `this.search` is bound to `this` in the constructor, so `this._search()` should be a valid reference in here. And the inner function is an arrow func, so the `this` binding should work there, too. Maybe give it another try using `this` in place of `server`?

lib/recommendation-server.js
```diff
@@ -82,17 +89,36 @@ RecommendationServer.prototype = {
...
     if (!payload.query || payload.query.length > this.maxChars) {
       return;
     }
-    this._search(payload.query).then(data => {
-      this.events.publish('recommendation', data);
+    server._search(payload.query).then(data => {
+      let resultClass = server.getResultClass(data);
```
jaredhirsch: Since we're inside a bound function, you don't need to bind `getResultClass` in the constructor. `this.getResultClass` should work just fine in here.

lib/ui/recommendation.js
```diff
@@ -123,21 +123,25 @@ Recommendation.prototype = {
     this.events.publish('recommendation-mousemove', evt);
   },
   show: function(data) {
+    let payload = data.payload;
+    let ResultClass = data.resultClass;
```
jaredhirsch: let's avoid initial uppercase letters unless it's a constructor function or an all-caps constant

chuckharmston: It is a constructor 👍  Below: `new ResultClass(payload`

lib/ui/recommendation.js
```diff
@@ -123,21 +123,25 @@ Recommendation.prototype = {
...
+
     if (this.timeout) {
```
jaredhirsch: Could you rename `this.timeout` to `this.hideTimeout` as part of this patch?

lib/ui/recommendation.js
```diff
@@ -150,7 +154,7 @@ Recommendation.prototype = {
           this.el.appendChild(recommendation);
         }
         this.el.collapsed = false;
-        this.data = data;
+        this.data = payload;
```
jaredhirsch: I'd prefer to keep `this.data` and get rid of `this.prev`. Is there any reason we need to keep two copies of the data?

jaredhirsch: If we just don't null out `this.data` on line 163 and 174, we should be able to get rid of `this.prev`, I think?

lib/ui/recommendation.js
```diff
@@ -123,21 +123,25 @@ Recommendation.prototype = {
...
+
     if (this.timeout) {
       this.win.clearTimeout(this.timeout);
       this.timeout = null;
     }
 
     // Don't re-render the recommendation if the content hasn't changed since
     // the previous recommendation.
-    if (this.prev && this.el && this.el.firstChild && data.result.url === this.prev.result.url) {
+    if (this.prev && this.el && this.el.firstChild &&
```
jaredhirsch: Rather than check for `this.el` here and at line 149 below, just check once, at the very top of the function (before the timeout check).

lib/ui/recommendation.js
```diff
@@ -123,21 +123,25 @@ Recommendation.prototype = {
...
+        payload.result.url === this.prev.result.url) {
       this.el.collapsed = false;
       this.events.publish('recommendation-shown', this);
       return;
     }
 
-    this.prev = data;
-    const recommendation = this.row = new this.RecommendationRow(data, {
+    this.prev = payload;
+    const recommendation = this.row = new ResultClass(payload, {
```
jaredhirsch: Why do we need to do an `if (recommendation)` check at line 150 below? The constructor should always return an object, which is truthy

lib/ui/results.js
```diff
@@ -0,0 +1,338 @@
...
+    const popupWidth = Math.floor(
+      result.doc.getElementById('PopupAutoCompleteRichResult').width);
+    const available = popupWidth - MAGIC;
+    return available >= MIN ? available : MIN;
+  }
+};
+
+
+
+function TLDResult(data, opts) {
```
jaredhirsch: Let's stick with one class per file, and make sure the name of the file, the name of the class, and `EXPORTED_SYMBOLS` all match.

lib/ui/results.js
```diff
@@ -0,0 +1,338 @@
...
+  return result._get('enhancements.wikipedia.url');
+};
+WikipediaResult.prototype.getTitleText = function(result) {
+  let suffix = / \- Wikipedia, the free encyclopedia$/;
+  return result._get('enhancements.wikipedia.title').replace(suffix, '');
+};
+
+
+
+function createElement(doc, elementName, id) {
```
jaredhirsch: Seems weird to have this function floating around here. Move it to the BaseResult prototype

lib/ui/results.js
chuckharmston: I agree. How would you recommend doing the imports? Is it okay to just import them into this file, or would you prefer that the `resultsClass` constant is in the main module where the majority of our imports happen?

jaredhirsch: Constructor dependency injection seems especially awkward with a hierarchy of UI classes, so feel free to just `Cu.import` the `base-result.js` module inside the subclasses.

jaredhirsch:
It looks like `result` is passed to every single rendering method. It really belongs on `this`, since it does belong to the object. Is this because of some problems with binding the value of `this` somewhere? If you use the this-binding approach I mentioned [in another comment](https://github.com/mozilla/universal-search/pull/211#discussion_r63229821), what seems to be broken?

lib/ui/results.js
jaredhirsch: > I agree. How would you recommend doing the imports? Is it okay to just import them into this file, or would you prefer that the resultsClass constant is in the main module where the majority of our imports happen?

Let's do one UI class per file: base-result.js, tld-result.js, wikipedia-result.js. To inherit from the BaseResult, just Cu.import the base-result.js file directly. Let EXPORTED_SYMBOLS be just a single class per file: BaseResult, WikipediaResult, or TLDResult.

lib/recommendation-server.js
jaredhirsch: I don't think that's relevant here, since it's just a data object. Nobody's modifying its prototype.

lib/ui/recommendation.js
jaredhirsch: I don't think the recommendation-server should have knowledge of the UI classes. Better to have this recommendation class pick a UI class based on properties of the data returned by the server.

lib/recommendation-server.js
jaredhirsch: Per my other comment, I actually think this should be moved to the recommendation.js class, since that class has the responsibility of picking a UI class that fits the data. The recommendation-server should just abstract out the service endpoint and return nicely formatted data (IMO)

lib/recommendation-server.js
```diff
@@ -29,10 +29,13 @@ function RecommendationServer(opts) {
   this.xhr = opts.xhr;
   this.console = opts.console;
   this.prefs = opts.prefs;
+  this.resultClasses = opts.resultClasses;
   this.urlbarPrefs = opts.urlbarPrefs;
   this.server = this.prefs.getPrefType('server') ?
                 this.prefs.getCharPref('server') :
                 'https://universalsearch.testpilot.firefox.com';
+
+  this.getResultClass = this.getResultClass.bind(this);
```
jaredhirsch: This line and `opts.resultClasses` / `this.resultClasses` should also go away when `getResultClass` is moved

jaredhirsch:
@chuckharmston After looking at the three result UI classes for a while (BaseResult, WikipediaResult, TLDResult), I don't think the subclasses are necessary. The variation among the subclasses is data, not code, except for the `getImageWidth` / `getImageHeight` methods in the Wikipedia class, and those two methods could just be moved to the base class.

Rather than generalize now for the sake of the possible future variations introduced by movie, product, or weather cards, I'd prefer to keep this code as simple as possible, and only generalize when we have a specific, concrete requirement that forces us to add code. (That is, let's generalize in the movie card PR, and not sooner.)

I'd suggest one result UI class, and TLD or Wikipedia-specific transformers that pass a data blob into the BaseResult constructor:

``` js
// tld type
{ type: 'tld',
  imageUrl: result._get('enhancements.tld'),
  imageWidth: 64, // just hard-code img dimensions here
  imageHeight: 64
}

// wikipedia type
{ type: 'wikipedia',
  faviconUrl: 'chrome://universalsearch-skin/content/wikipedia-dark.svg',
  imageUrl: result._get('enhancements.wikipedia.image.url'),
  url: result._get('enhancements.wikipedia.url'),
  titleText: title minus the wikipedia bits (see getTitleText).
}
```

Happy to discuss further :-)

lib/recommendation-server.js
```diff
@@ -82,17 +89,36 @@ RecommendationServer.prototype = {
...
+    let server = this;
     if (!payload.query || payload.query.length > this.maxChars) {
       return;
     }
-    this._search(payload.query).then(data => {
-      this.events.publish('recommendation', data);
+    server._search(payload.query).then(data => {
+      let resultClass = server.getResultClass(data);
+      if (resultClass) {
+        server.events.publish('recommendation', {
```
jaredhirsch: This might be a good opportunity to make this a more self-documenting event name. Maybe `'recommendation-data'`?

jaredhirsch:
Forgot to add: great job! This was clearly a ton of work, and it looks great overall. We'll get it landed in another pass, I bet

jaredhirsch:
Whoops, looks like I forgot to reassign this on Friday :-\

lib/recommendation-server.js
chuckharmston: ☑️

lib/recommendation-server.js
chuckharmston, line 93: ☑️

lib/recommendation-server.js
chuckharmston: ☑️

lib/ui/recommendation.js
chuckharmston: ☑️

lib/ui/recommendation.js
chuckharmston: ☑️

lib/ui/recommendation.js
chuckharmston: ☑️

lib/ui/recommendation.js
chuckharmston: I allowed for the possibility for the results classes themselves to decide, based on the data, to not recommend something. That's a little duplicitous though, I'll remove it.

lib/ui/results.js
chuckharmston: ☑️

lib/ui/results.js
chuckharmston: ☑️

lib/ui/recommendation.js
chuckharmston: ☑️

lib/ui/recommendation.js
chuckharmston: ☑️

lib/ui/recommendation.js
chuckharmston: ✔️

lib/ui/recommendation.js
chuckharmston: ✔️

chuckharmston:
@6a68: as we discussed a few weeks ago, the inheritance chain with these updates looks like this:

```
                            TLDResult
             SimpleResult <
BaseResult <                WikipediaResult
             MovieResult
```

Where:
- `BaseResult` manages construction, options, and defines the basic interface (i.e. `render`)
  - `SimpleResult` contains the logic and construction for results that conform to the basic favicon/image/text/url format. It abstracts a method responsible for negotiating bits of data that fit into those various slots.
    - `TLDResult` translates data -> display for TLDs.
    - `WikipediaResult` translates data -> display for Wikipedia.
  - The future `MovieResult` uses a quite different layout, and will subclass `BaseResult`.

This is ready for rereview.

jaredhirsch:
@chuckharmston Hey, it seems like you may have missed my [last comment](https://github.com/mozilla/universal-search/pull/211#issuecomment-219162988) from the previous round. I don't see the need for a complex hierarchy of UI classes: the two items are rendered in the same way, all that's needed is to slightly transform the data. Let's talk about this tomorrow.

jaredhirsch:
Closing per our discussion outside Github.

PR #226, opened 2016-05-17
Duplicate check

jaredhirsch:
@chuckharmston Debugged, fully working, and updated.

We are now waiting on two async things: the C++ controller polling code, which tells us what result URLs are about to hit the popup; and the recommendation xhr code, which tells us what the recommendation URL will be. We can't use promises to synchronize the two, because promises always wait a turn to resolve/reject, and the state of the world might change during that time.

Instead, we set a bit of JS state when the recommendation arrives, and when the C++ controller says it's done; a third function, `duplicateCheck`, is called in both cases, so that we run the check as soon as both pieces of data are available.

The duplicate check compares the URLs via the Gecko `nsIURI` object, which does some nice normalization and canonicalization for us.

### DIFF
--- a/lib/ui/recommendation.js
+++ b/lib/ui/recommendation.js
@@ -4,6 +4,10 @@
 
 const EXPORTED_SYMBOLS = ['Recommendation'];
 const RECOMMENDATION_ID = 'universal-search-recommendation';
+// borrowed from urlbarBindings.xml, _parseActionUrl method
+const MOZ_ACTION_REGEX = /^moz-action:([^,]+),(.*)$/;
+
+const Ci = Components.interfaces;
 
 function Recommendation(opts) {
   /*
@@ -21,6 +25,11 @@ function Recommendation(opts) {
   this.mouseMoveTimeout = null;
   this.prev = null;
   this.row = null;
+  this.searchResults = null;
+  // this.recData is used for duplicate checking, but sort of overlaps with
+  // this.prev / this.data. Consider just storing the data in one place.
+  this.recData = null;
+  this.searchController = null;
 
   Object.defineProperty(this, 'isHighlighted', {
     get: () => {
@@ -40,8 +49,11 @@ function Recommendation(opts) {
 
   this.navigate = this.navigate.bind(this);
   this.show = this.show.bind(this);
+  this.onData = this.onData.bind(this);
   this.hide = this.hide.bind(this);
   this._pollForElement = this._pollForElement.bind(this);
+  this.pollController = this.pollController.bind(this);
+  this._controllerCheck = this._controllerCheck.bind(this);
   this.onMouseEnter = this.onMouseEnter.bind(this);
   this.onMouseMove = this.onMouseMove.bind(this);
   this.onMouseLeave = this.onMouseLeave.bind(this);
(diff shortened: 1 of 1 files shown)


lib/ui/recommendation.js
```diff
@@ -187,5 +197,125 @@ Recommendation.prototype = {
...
+
+        // 'moz-action' URLs (switch to tab, or search suggestion) have an empty
+        // string asciiHost. Don't bother trying to compare domains in this case.
+        if (!resultURI.asciiHost) {
+          return;
+        }
+
+        // The eTLD service should give us a good, cheap guess at the domain.
+        if (this.eTLD.getBaseDomain(resultURI) === this.eTLD.getBaseDomain(recommendationURI)) {
+          isDuplicate = true;
```
jaredhirsch: derp. we need to bail on the forEach loop as soon as we see a duplicate

lib/ui/recommendation.js
```diff
@@ -187,5 +197,125 @@ Recommendation.prototype = {
...
+    // Each time the urlbar changes, use a tight setTimeout loop to poll the
+    // C++ controller for results.
+    this.ctrl = Components.classes["@mozilla.org/autocomplete/controller;1"].
+                getService(Components.interfaces.nsIAutoCompleteController);
+
+    this._controllerCheck();
+  },
+  _controllerCheck: function() {
+    // searchStatus 4 means search succeeded and is done
+    if (this.ctrl.searchStatus == 4) {
```
jaredhirsch: probably better to use the [`Ci.nsIAutoCompleteController` constants](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/autocomplete/nsIAutoCompleteController.idl#15-18) than random-looking integers

lib/ui/recommendation.js
```diff
@@ -59,9 +65,10 @@ Recommendation.prototype = {
       this.events.publish('recommendation-created');
     });
 
-    this.events.subscribe('recommendation', this.show);
+    this.events.subscribe('recommendation', this.onData);
     this.events.subscribe('enter-key', this.navigate);
     this.events.subscribe('urlbar-change', this.hide);
+    this.events.subscribe('urlbar-change', this._pollController);
```
jaredhirsch: Seems weird to be exposing an underscored "private-ish" method to external code. Better to remove the underscore

jaredhirsch:
I'll just land this myself after taking a fresh look in the morning.

Let's celebrate with a bizarre gif :tada:

lib/ui/recommendation.js
```diff
@@ -59,15 +73,33 @@ Recommendation.prototype = {
...
+    // optimistically trigger the duplicate check.
+    this.events.subscribe('urlbar-change', this.pollController);
+
+    // When the urlbar changes, the recommendation becomes stale, so hide it.
     this.events.subscribe('urlbar-change', this.hide);
+
+    // When the popup is about to close, hide the recommendation.
     this.events.subscribe('before-popup-hide', this.hide);
+
+    this.ctrl = Components.classes["@mozilla.org/autocomplete/controller;1"].
```
jaredhirsch: pretty cryptic variable name

lib/ui/recommendation.js
```diff
@@ -187,5 +221,164 @@ Recommendation.prototype = {
...
+          continue;
+        }
+      }
+
+      // Finally, check for duplicates.
+      // For TLD results, the eTLD service should give us a good, cheap guess
+      // at the domain.
+      // For wikipedia results, use nsIURI.equals() to check for URL equality.
+      // If a duplicate is found, immediately exit; else, show the
+      // recommendation.
```
jaredhirsch: I think this code is excessively documented. Maybe just document surprising / weird things (like the moz-action stuff). There's no need to say in English exactly what the code says in JavaScript.

lib/ui/recommendation.js
```diff
@@ -187,5 +221,164 @@ Recommendation.prototype = {
...
+      if (type === 'tld' && this.eTLD.getBaseDomain(resultUri) === this.eTLD.getBaseDomain(recommendationUri) ||
+          type === 'wikipedia' && resultUri.equals(recommendationUri)) {
+        return;
+      }
+    }
+
+    this.show(this.recData);
+  },
+  _isMozAction: function(/* nsIURI */ uri) {
+    return MOZ_ACTION_REGEX.test(uri.asciiSpec);
```
jaredhirsch, line 336: Do we really need a separate function for a short one-liner?

lib/ui/recommendation.js
jaredhirsch: fixed

lib/ui/recommendation.js
jaredhirsch, line 336: It's fine, keeps that stanza in the duplicateCheck function easier to skim.

jaredhirsch:
Working for me in Nightly and Release channels. Detects switch-to-tab and heuristic domain duplicates. Merging.

PR #230, opened 2016-05-20
Use 'popuphidden' instead of 'popuphiding' to detect XUL popup state

### LINKED ISSUE #164: Search suggestions and Places results not rendering after upgrading Nightly
The new popup UI rolled out to developer edition today, and with our addon installed you no longer see any search suggestions or results from Places.

screen shot 2016-04-29 at 7 28 40 pm

### LINKED ISSUE #187: Use XBL handlers to detect XBL events
The `popuphiding` handler defined in `autocomplete.xml` seems to be overwritten by the `popuphiding` handler registered in JS-DOM in popup.js.

While debugging #164, it occurred to me that, if we move the handler to our XBL binding, then (per XBL's binding inheritance rules) both our handler and the parent binding's handlers would execute.

I think this would have prevented #164, even if that bug was ultimately a transient nightly bug.

Related: the fix for #138 involved some hacky state setting the first time the popup closes. If we move the popuphiding binding over to XBL, I bet we can remove the `onFirstPopupOpen` method introduced in https://github.com/mozilla/universal-search/commit/3afba285082dc1f7404513fa34622faa70a9f6b8.

Note that the bug for #164 was fixed if we manually updated `gURLBar.maxRows` from -1 to 10 after the first time the popup closed. Same workaround as that used in #138.

---

The work required to test this is to replace the `addEventListener` calls [inside popup.js](https://github.com/mozilla/universal-search/blob/master/lib/ui/popup.js#L27-L29) with XBL one-liner handlers like those [we already use](https://github.com/mozilla/universal-search/blob/master/chrome/content/binding.xml#L49-L51) to catch some events. Basically, the event gets fired on the XBL side, then the XBL handler calls the JS handler via the app global `window.universalSearch`, which is shared between XBL and JS-DOM.

jaredhirsch:
It turns out[1] that DOM Level 2 `addEventListener` handlers, when added
using JS, override XBL handlers for the same events. Nothing seems to
listen for the 'popuphidden' event that's fired just after
'popuphiding', and for our purposes (starting a metrics interaction
event, hiding the recommendation, resetting interaction state in the
highlight manager), either event works fine.

Update the pubsub event from 'before-popup-hide' to 'after-popup-hide',
so that our event names reflect the browser's reality as closely as
possible.

Fixes #164, closes #187.

[1] http://mb.eschew.org/15#sub_15.2.7.1

### DIFF
--- a/lib/metrics.js
+++ b/lib/metrics.js
@@ -37,14 +37,14 @@ Metrics.prototype = {
     this.events.subscribe('urlbar-change', this.interactionStart);
 
     // The interaction ends when the popup closes.
-    this.events.subscribe('before-popup-hide', this.interactionEnd);
+    this.events.subscribe('after-popup-hide', this.interactionEnd);
   },
   destroy: function() {
     this.events.unsubscribe('before-popup-show', this.interactionStart);
     this.events.unsubscribe('before-keyboard-navigate', this.onKeyNavigate);
     this.events.unsubscribe('before-click-navigate', this.onClickNavigate);
     this.events.unsubscribe('urlbar-change', this.interactionStart);
-    this.events.unsubscribe('before-popup-hide', this.interactionEnd);
+    this.events.unsubscribe('after-popup-hide', this.interactionEnd);
   },
   interactionStart: function() {
     // Initialize default values for the case where the popup closes without a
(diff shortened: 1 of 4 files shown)


jaredhirsch:
I'm no longer able to reproduce the underlying bug without this patch applied, so I can't verify that this patch fixes the bug. However, I was able to repro the bug last night, and this patch did fix it. Everything else seems to work as usual, so, merging.

PR #236, opened 2016-05-24
add unusual field to metrics docs

clouserw:


clouserw:
This is just documentation updates for our new column, as discussed in https://github.com/mozilla/universal-search-recommendation/issues/138

PR #240, opened 2016-05-26
movie card (WIP)

jaredhirsch:
@chuckharmston Not totally done yet, but I'd like to get your feedback before you head out for the weekend.

Still needs styling, I'm not sure the state management is totally correct (this.row, this.data, this.rawData), and there's a bug where stale recommendations get re-shown. If you see other bugs, let me know. 

Comments / questions welcome. Might be slightly more readable per-commit than as a summary diff.

### DIFF
--- a/lib/ui/recommendation.js
+++ b/lib/ui/recommendation.js
@@ -21,7 +21,8 @@ function Recommendation(opts) {
   this.RecommendationRow = opts.RecommendationRow;
   this.transforms = {
     tld: opts.tldTransform,
-    wikipedia: opts.wikipediaTransform
+    wikipedia: opts.wikipediaTransform,
+    movie: opts.movieTransform
   };
 
   this.el = null;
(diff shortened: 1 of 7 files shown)


lib/ui/recommendation.js
```diff
@@ -263,22 +288,31 @@ Recommendation.prototype = {
...
     // a wikipedia recommendation is a duplicate if a result has the same URL;
     // a top-level domain recommendation is a duplicate if any result is a page
     // from that domain (we only want to show new domains).
+    if (!this.searchResults || !this.rawData) {
+      return;
+    }
 
-    if (!this.searchResults || !this.recData) {
+    let type = this.getResultType(this.rawData);
+    this.data = this.transforms[type].transform(this.rawData);
```
jaredhirsch: what if we wait to assign this.data until the dupe check inside show(), at line 172? we could pass the data to show() as a parameter. should fix the stale recommendation bug.

lib/ui/recommendation.js
```diff
@@ -324,13 +358,13 @@ Recommendation.prototype = {
...
-      if (type === 'tld' && this.eTLD.getBaseDomain(resultUri) === this.eTLD.getBaseDomain(recommendationUri) ||
-          type === 'wikipedia' && resultUri.equals(recommendationUri)) {
+      if (this.data.type === 'tld' && this.eTLD.getBaseDomain(resultUri) === this.eTLD.getBaseDomain(recommendationUri) ||
+          this.data.type === 'wikipedia' && resultUri.equals(recommendationUri)) {
         return;
       }
     }
 
-    this.show(this.recData);
+    this.show();
```
jaredhirsch: so, right here, pass in data.

jaredhirsch:
closing, reopening as WIP.

PR #244, opened 2016-06-01
agent field is broken out in redshift

clouserw:


jaredhirsch:
@clouserw @chuckharmston what's the status of this other PR? I'm guessing we do not need it?

clouserw:
I'm gonna land it anyway, since this is documenting something that already exists

Issue #241, opened 2016-05-27
Throw away old recommendation on browser resize


jaredhirsch:
We currently save the last-shown recommendation, just in case changing one char brings the same data back from the server. It's a nice performance tweak.

However, if we show a recommendation, then the window is resized, then the same recommendation is re-shown, the layout of the recommendation isn't recalculated (probably a XUL bug).

I think the simplest workaround would be, when reusing a recommendation row, reseat it in the DOM by replacing it with itself: `rec.parentNode.replaceChild(rec, rec)` (or something like that).

A slightly more complex workaround would be to listen for window resize events (using `addEventListener` might clobber XBL resize handlers, so it's better to just look for existing nsIObserver notifications, or fancy SDK libraries), and if a resize event occurs, just throw out the old recommendation. I think the first approach is better, if it works, as it lets us retain the performance optimization and fixes the bug.

PR #260, opened 2016-07-13
Wip movie card

jaredhirsch:
reviewing my own code to try to find the remaining minor bugs:
- movie cards seem to render a bit slowly
- dupe checking doesn't work for identical movie card imdb URLs

Also need to do some cleanup and ensure the commits make sense individually / as a group.

### DIFF
--- a/chrome/skin/style.css
+++ b/chrome/skin/style.css
@@ -26,9 +26,9 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 
 #universal-search-recommendation {
   border-bottom: none;
+  display: flex;
   padding: 8px 15px 8px 9px;
   width: 100%;
-  display: flex;
 }
 
 #universal-search-recommendation[collapsed="true"] {
@@ -64,7 +64,8 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 }
 
 #universal-search-recommendation-top-content {
-    -moz-box-pack: justify;
+  -moz-box-flex: 1;
+  -moz-box-pack: justify;
 }
 
 #universal-search-recommendation-letterbox {
@@ -112,22 +113,7 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 #universal-search-recommendation-content {
   flex-grow: 1;
   min-width: 0;
-}
-
-#universal-search-recommendation-details,
-#universal-search-recommendation-description {
-  color: GrayText;
-  width:100%;
-}
-
-#universal-search-recommendation-title,
-#universal-search-recommendation-url {
-  color: -moz-FieldText;
-  color: FieldText;
-/*  margin: 0; */
-  overflow: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
+  -moz-box-align: stretch;
 }
 
 #universal-search-recommendation-details,
@@ -135,10 +121,16 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
   color: GrayText;
   -moz-box-align: start;
   text-overflow: ellipsis;
+  width:100%;
 }
 
-#universal-search-recommendation-details {
-  width: 100%;
+#universal-search-recommendation-title,
+#universal-search-recommendation-url {
+  color: -moz-FieldText;
+  color: FieldText;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 }
 
 #universal-search-recommendation-url {
@@ -155,10 +147,6 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 #universal-search-recommendation-label {
   color: GrayText;
   cursor: default;
-  /*
-  margin: 0;
-  text-align: right;
-  */
   -moz-user-select: none;
   user-select: none;
 }
@@ -193,13 +181,4 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 }
 #universal-search-recommendation.highlight #universal-search-recommendation-keyimage {
   z-index: 1;
-}
-
-#universal-search-recommendation-content {
-  -moz-box-align: stretch;
-}
-
-#universal-search-recommendation-top-content {
-  -moz-box-flex: 1;
-  -moz-box-pack: justify;
 }
--- a/chrome/skin/style.css
+++ b/chrome/skin/style.css
@@ -26,7 +26,6 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 
 #universal-search-recommendation {
   border-bottom: none;
-  display: flex;
   padding: 8px 15px 8px 9px;
   width: 100%;
 }
--- a/chrome/skin/style.css
+++ b/chrome/skin/style.css
@@ -28,6 +28,7 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
   border-bottom: none;
   padding: 8px 15px 8px 9px;
   width: 100%;
+  display: flex;
 }
 
 #universal-search-recommendation[collapsed="true"] {
@@ -36,7 +37,7 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 
 #universal-search-recommendation-container {
   display: flex;
-  width: 100%;
+  flex: 1;
 }
 
 #universal-search-recommendation-favicon {
@@ -60,6 +61,10 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 #universal-search-recommendation-keyimage {
   position: relative;
   z-index: -1;
+}
+
+#universal-search-recommendation-top-content {
+    -moz-box-pack: justify;
 }
 
 #universal-search-recommendation-letterbox {
@@ -112,14 +117,14 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 #universal-search-recommendation-details,
 #universal-search-recommendation-description {
   color: GrayText;
+  width:100%;
 }
 
 #universal-search-recommendation-title,
 #universal-search-recommendation-url {
   color: -moz-FieldText;
   color: FieldText;
-  display: block;
-  margin: 0;
+/*  margin: 0; */
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
@@ -130,7 +135,10 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
   color: GrayText;
   -moz-box-align: start;
   text-overflow: ellipsis;
-  margin-left: 0;
+}
+
+#universal-search-recommendation-details {
+  width: 100%;
 }
 
 #universal-search-recommendation-url {
@@ -147,9 +155,10 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 #universal-search-recommendation-label {
   color: GrayText;
   cursor: default;
-  flex-grow: 1;
-  margin-left: 8px;
+  /*
+  margin: 0;
   text-align: right;
+  */
   -moz-user-select: none;
   user-select: none;
 }
@@ -184,4 +193,13 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 }
 #universal-search-recommendation.highlight #universal-search-recommendation-keyimage {
   z-index: 1;
+}
+
+#universal-search-recommendation-content {
+  -moz-box-align: stretch;
+}
+
+#universal-search-recommendation-top-content {
+  -moz-box-flex: 1;
+  -moz-box-pack: justify;
 }
--- a/chrome/skin/style.css
+++ b/chrome/skin/style.css
@@ -110,6 +110,11 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
   min-width: 0;
 }
 
+#universal-search-recommendation-details,
+#universal-search-recommendation-description {
+  color: GrayText;
+}
+
 #universal-search-recommendation-title,
 #universal-search-recommendation-url {
   color: -moz-FieldText;
@@ -119,6 +124,14 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
+}
+
+#universal-search-recommendation-details,
+#universal-search-recommendation-description {
+  color: GrayText;
+  -moz-box-align: start;
+  text-overflow: ellipsis;
+  margin-left: 0;
 }
 
 #universal-search-recommendation-url {
@@ -133,7 +146,7 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 }
 
 #universal-search-recommendation-label {
-  color: #BBB;
+  color: GrayText;
   cursor: default;
   flex-grow: 1;
   margin-left: 8px;
@@ -160,7 +173,9 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 #universal-search-recommendation.highlight #universal-search-recommendation-url {
   color: HighlightText;
 }
-#universal-search-recommendation.highlight #universal-search-recommendation-label {
+#universal-search-recommendation.highlight #universal-search-recommendation-label,
+#universal-search-recommendation.highlight #universal-search-recommendation-details,
+#universal-search-recommendation.highlight #universal-search-recommendation-description {
   color: HighlightText;
   opacity: 0.67;
 }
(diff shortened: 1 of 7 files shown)


chrome/skin/style.css
```diff
@@ -26,9 +26,9 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
 
 #universal-search-recommendation {
   border-bottom: none;
-  display: flex;
   padding: 8px 15px 8px 9px;
   width: 100%;
+  display: flex;
```
jaredhirsch: just put display:flex back where it was

lib/ui/recommendation.js
```diff
@@ -167,14 +169,21 @@ Recommendation.prototype = {
       return;
     }
 
+    this.win.console.log('about to dupe check inside show().');
+
     // Reuse the recommendation if it's still in the DOM and the content hasn't
     // changed since the previous recommendation.
-    if (this.el.firstChild && this.data && this.row.url === this.data.url) {
+    if (this.el.firstChild && data && this.row &&
+        this.row.url === data.url) {
```
6a68: why would `this.row.url !== data.url`? set the debugger and find out.

6a68: the `data` vs `this.data` distinction seems potentially tricky.

lib/ui/recommendation.js
```diff
@@ -206,27 +215,31 @@ Recommendation.prototype = {
...
-  },
-  hideImmediate: function() {
-    if (this.hideTimeout) {
-      this.win.clearTimeout(this.hideTimeout);
-      this.hideTimeout = null;
+  hide: function(isImmediate) {
+    this.win.console.log('hide: ', Date.now());
+    this.searchResults = null;
+    this.data = null;
+    this.rawData = null;
```
6a68: this.data vs this.rawData?

lib/ui/recommendation.js
```diff
@@ -300,10 +313,11 @@ Recommendation.prototype = {
...
 
     let type = this.getResultType(this.rawData);
-    this.data = this.transforms[type].transform(this.rawData);
+    let data = this.transforms[type].transform(this.rawData);
     // If the data was malformed, immediately hide any existing recommendation.
-    if (!this.data) {
-      this.hideImmediate();
+    if (!data) {
+      this.win.console.log('no data returned by transform, hiding immediately');
+      this.hide(true);
```
jaredhirsch, line 328: I like hideImmediate better than hide(true), let's switch back to that.

chrome/skin/style.css
```diff
@@ -166,15 +181,25 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
...
   z-index: 1;
 }
+
+#universal-search-recommendation-content {
+  -moz-box-align: stretch;
+}
+
+#universal-search-recommendation-top-content {
+  -moz-box-flex: 1;
+  -moz-box-pack: justify;
```
jaredhirsch: this rule is a duplicate of the rule above. let's merge this into the one above for `#top-content`.

chrome/skin/style.css
```diff
@@ -139,11 +153,12 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
...
 #universal-search-recommendation-label {
-  color: #BBB;
+  color: GrayText;
   cursor: default;
-  flex-grow: 1;
-  margin-left: 8px;
+  /*
+  margin: 0;
   text-align: right;
+  */
```
jaredhirsch: obviously clean all this up

lib/movie-transform.js
```diff
@@ -0,0 +1,65 @@
...
+      valid = data.enhancements.movie.title &&
+              data.enhancements.movie.imdb_url &&
+              data.enhancements.favicon.url &&
+              data.enhancements.favicon.color &&
+              data.enhancements.movie.poster &&
+              data.enhancements.movie.rating.imdb &&
+              data.enhancements.movie.runtime &&
+              data.enhancements.movie.genre;
+    } catch (ex) {}
+    return !!valid;
```
jaredhirsch: Seems clearer to just do `let valid = false;` and return `valid` at the end.

lib/tld-transform.js
```diff
@@ -0,0 +1,55 @@
...
+        width: IMAGE_DIMENSION
+      }
+    };
+    return result;
+  },
+  isValid: function(data) {
+    let valid;
+    // Just try to access every required key. If one is missing, the thrown Error
+    // will tell us it's an invalid data object. For optional keys, just check
+    // that the parent key exists (if no other check touches the parent).
```
jaredhirsch, line 45: Put this comment in all of the isValid functions, and use the more direct `let valid = false;` everywhere.

lib/wikipedia-transform.js
```diff
@@ -0,0 +1,82 @@
...
+  transform: function(data) {
+    if (!this.isValid(data)) {
+      return;
+    }
+    let dimensions = {};
+    if (data.enhancements.wikipedia.image.height &&
+        data.enhancements.wikipedia.image.width) {
+      dimensions = this.calculateDimensions(data);
+    }
+    // NOTE: using the full keys repeatedly to aid grepping.
```
jaredhirsch, line 29: Put this note everywhere or nowhere :-)

lib/wikipedia-transform.js
```diff
@@ -0,0 +1,82 @@
...
+      };
+    }
+
+    return {
+      width: IMAGE_DIMENSION,
+      height: height / width * IMAGE_DIMENSION
+    };
+  },
+  isValid: function(data) {
+    let valid;
```
jaredhirsch: again, use `let valid = false;`

lib/wikipedia-transform.js
```diff
@@ -0,0 +1,82 @@
...
+      type: 'wikipedia',
+      favicon: {
+        url: 'chrome://universalsearch-skin/content/wikipedia-dark.svg',
+        color: data.enhancements.favicon.color
+      },
+      title: data.enhancements.wikipedia.title,
+      url: data.enhancements.wikipedia.url,
+      keyImage: {
+        url: data.enhancements.wikipedia.image.url,
+        height: dimensions.height,
```
6a68: this could be `undefined`. does the rendering code check truthiness, or will it render with size `'undefinedpx'`?

lib/ui/recommendation.js
```diff
@@ -324,13 +378,13 @@ Recommendation.prototype = {
       // at the domain.
       // For wikipedia results, use nsIURI.equals() to check for URL equality.
       // If a duplicate is found, immediately exit.
-      if (type === 'tld' && this.eTLD.getBaseDomain(resultUri) === this.eTLD.getBaseDomain(recommendationUri) ||
-          type === 'wikipedia' && resultUri.equals(recommendationUri)) {
+      if (data.type === 'tld' && this.eTLD.getBaseDomain(resultUri) === this.eTLD.getBaseDomain(recommendationUri) ||
+          data.type === 'wikipedia' && resultUri.equals(recommendationUri)) {
         return;
```
jaredhirsch, line 386: Need to add the movie type here. Nice, simple explanation for why dupe checking fails to work ;-)

lib/wikipedia-transform.js
6a68: It assumes truthiness and checks if the image URL exists.

So, the way to avoid rendering some broken junk is to append the keyImage after `result` is created, and only if the url is set.

lib/wikipedia-transform.js
```diff
@@ -0,0 +1,82 @@
...
+  This class is stateless, so a singleton instance is created and exported.
+  */
+}
+WikipediaTransform.prototype = {
+  transform: function(data) {
+    if (!this.isValid(data)) {
+      return;
+    }
+    let dimensions = {};
+    if (data.enhancements.wikipedia.image.height &&
```
jaredhirsch, line 25: Is `wikipedia.image` always defined?

jaredhirsch, line 25: Yes, because `this.isValid` bails if `wikipedia.image` is falsy. Awesome.

lib/ui/recommendation.js
```diff
@@ -19,16 +19,21 @@ function Recommendation(opts) {
...
 
   this.el = null;
-  this.timeout = null;
+  this.hideTimeout = null;
   this.mouseMoveTimeout = null;
-  this.prev = null;
+  // this.rawData is the raw response data from the server.
+  this.rawData = null;
+  // this.data is the data after its type is inferred and it is transformed.
+  this.data = null;
```
jaredhirsch, line 34: Actually, these docs for `this.data` vs `this.rawData` are great.

lib/ui/recommendation.js
```diff
@@ -19,16 +19,21 @@ function Recommendation(opts) {
...
   this.RecommendationRow = opts.RecommendationRow;
+  this.transforms = {
+    tld: opts.tldTransform,
+    wikipedia: opts.wikipediaTransform,
+    movie: opts.movieTransform
+  };
 
   this.el = null;
-  this.timeout = null;
+  this.hideTimeout = null;
```
jaredhirsch: Add a comment in the final PR that this name change was just to clarify the purpose of the timeout.

jaredhirsch: Or maybe get rid of this change. Wasn't there a moment when I tried using two timeouts, hence, needed to distinguish between them?

lib/ui/recommendation.js
```diff
@@ -154,53 +159,92 @@ Recommendation.prototype = {
...
+    this.searchResults = null;
+    this.data = null;
+    this.rawData = null;
+
+    if (!this.el) {
+      return;
+    }
+    // this.hide is called by some pubsub signals, which pass in truthy
+    // arguments, but we only want to immediately hide if the argument was
+    // the boolean 'true'.
```
jaredhirsch: Move this documentation to the top of the function

lib/ui/recommendation.js
```diff
@@ -154,53 +159,92 @@ Recommendation.prototype = {
...
-    // Don't re-render the recommendation if the content hasn't changed since
-    // the previous recommendation.
-    if (this.prev && this.el && this.el.firstChild && data.result.url === this.prev.result.url) {
+    this.win.console.log('about to dupe check inside show().');
+
+    // Reuse the recommendation if it's still in the DOM and the content hasn't
+    // changed since the previous recommendation.
+    if (this.el.firstChild && data && this.row &&
+        this.row.url === data.url) {
       this.el.collapsed = false;
```
jaredhirsch, line 180: What about using the idea from the description for #241 here: instead of just unsetting `this.el.collapsed`, you should remove and reseat it in the XUL DOM to trigger a re-layout: `rec.parentNode.replaceChild(rec, rec)`?

lib/ui/recommendation.js
```diff
@@ -154,53 +159,92 @@ Recommendation.prototype = {
...
+      this.win.console.log('show found a dupe in the DOM, bailing.');
       return;
     }
 
-    this.prev = data;
-    const recommendation = this.row = new this.RecommendationRow(data, {
+    this.data = data;
+
+    this.win.console.log('not a dupe. this.row.url is',this.row && this.row.url,', and data.url is ',data.url);
+    this.row = new this.RecommendationRow(this.data, {
```
jaredhirsch, line 189: Do we need to `delete this.row` before creating a new row? Will the old row be properly GC'd?

jaredhirsch:
I wonder. What if we reused one recommendation, instead of continually creating new XUL nodes? Would the recommendation render faster?

lib/ui/recommendation.js
6a68: Nah, it's actually more readable, I think.

jaredhirsch:
Ah, nope. There's a code comment that mentions that the reused node's dimensions aren't recalculated; if the window was resized, the recommendation won't detect it. Yay XUL?

jaredhirsch:
lol, I meant to open this against my own fork. Ah well.

lib/ui/recommendation.js
jaredhirsch, line 180: Nope, see https://github.com/mozilla/universal-search/pull/260#issuecomment-232227271

chrome/skin/style.css
jaredhirsch: done

chrome/skin/style.css
```diff
@@ -116,17 +114,33 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
...
+  width:100%;
+}
+
 #universal-search-recommendation-title,
 #universal-search-recommendation-url {
   color: -moz-FieldText;
   color: FieldText;
-  display: block;
-  margin: 0;
+/*  margin: 0; */
```
jaredhirsch: deleted

chrome/skin/style.css
```diff
@@ -116,17 +114,33 @@ popupset#mainPopupSet panel#PopupAutoCompleteRichResult richlistbox {
...
+#universal-search-recommendation-details,
+#universal-search-recommendation-description {
+  color: GrayText;
+  -moz-box-align: start;
+  text-overflow: ellipsis;
+}
+
+#universal-search-recommendation-details {
+  width: 100%;
+}
```
jaredhirsch: Don't rely on position in the file to cascade changes. Try to define things as completely as possible all at once, so much easier to skim later.

jaredhirsch: Nope, messing with this breaks stuff. Just leave it as it is, as much as possible.

chrome/skin/style.css
jaredhirsch: done

chrome/skin/style.css
jaredhirsch: done

lib/movie-transform.js
jaredhirsch: done

lib/tld-transform.js
jaredhirsch, line 45: done

lib/wikipedia-transform.js
6a68: It looks like the server always returns url + height + width. We can just check for all three in all the validators.

lib/ui/recommendation-row.js
```diff
@@ -105,71 +81,49 @@ RecommendationRow.prototype = {
...
+    elem.setAttribute('height', keyImage.height);
+    elem.style.height = `${keyImage.height}px`;
+    elem.setAttribute('width', keyImage.width);
+    elem.style.width = `${keyImage.width}px`;
 
+    // Note: in both the resolve() and reject() callbacks, elem.parentNode
+    // is assumed to exist. This is correct, because `elem` is returned before
+    // the promise resolves, and the caller appends `elem` to the DOM in the
+    // same turn. This is one case where it's really nice that promise
+    // resolution requires waiting a turn :-)
```
jaredhirsch, line 113: Clarify this comment.

lib/ui/recommendation-row.js
```diff
@@ -198,76 +152,102 @@ RecommendationRow.prototype = {
...
     /*
-    Creates and returns a <vbox> element containing the recommendation's title
-    and URL.
+    Creates and returns three <hbox> elements containing:
+    - the title and the 'Recommended' label (so that description text flows
+      uninterruptedly beneath both)
+    - the details/description (if any)
+    - the url
     */
+    // TODO: remove the content ID styles from the CSS, replace with these.
```
jaredhirsch, line 161: What's this mean?

lib/ui/recommendation-row.js
```diff
@@ -336,7 +297,7 @@ RecommendationRow.prototype = {
     that there are at least 200px available for it.
     */
     const MIN = 200;
-    const MAGIC = 197;
+    const MAGIC = 107;
```
jaredhirsch, line 300: If this change has to be made, update the function's docs as well.

lib/ui/recommendation.js
jaredhirsch: yeah, let's revert and reduce the number of changes

jaredhirsch: done

lib/ui/recommendation.js
jaredhirsch, line 189: `¯\_(ツ)_/¯` let's not worry about this, our days here are numbered anyway

lib/ui/recommendation.js
jaredhirsch: done

lib/ui/recommendation.js
jaredhirsch, line 386: done

lib/wikipedia-transform.js
jaredhirsch, line 29: done

lib/wikipedia-transform.js
jaredhirsch: done

lib/ui/recommendation-row.js
jaredhirsch, line 113: done

lib/ui/recommendation-row.js
jaredhirsch, line 161: Seems like a half-forgotten refactoring. deleted

jaredhirsch:
Only remaining issue is that the min/max width of the recommendation needs tweaking. I'll open a for reals PR in the morning when that's done.

PR #261, opened 2016-07-14
200 movie card

jaredhirsch:
@chuckharmston R?

I had hoped to clean up the commit history before opening this for review, but I've run out of time. Probably best viewed as the unified diff; the individual commits are mostly grouped together well, but there are later bugfix commits that haven't been split and squashed. I'll clean up history before landing.

Points of interest:
- uses a set of transforms to ensure all required keys are in the data object before it's passed to the RecommendationRow to be rendered; this allowed a lot of conditional logic to be trimmed.
- we can trivially add A/B support for the movie card by tweaking the Recommendation's `getResultType` method, which assigns the type to a result before it's transformed. Line 213 of Recommendation.js might become something like `else if ('movie' in data.enhancements && variants.isMovieCardEnabled)`.
- tweak the XUL DOM bits to allow the movie description to flow under the 'Recommended' text. Most of these tweaks were bwinton's suggestions and I dimly understand them ;-)
- It turns out we can't reuse the movie card recommendations because XUL doesn't reflow elements that are removed / reinserted into the XUL DOM. This means that if you see a movie card, resize your browser window, then cause the same card to be shown, it'll have the wrong dimensions. To work around this, we discard the old recommendation at the end of the hide timeout.
- the movie cards are a bit slow to render. I think this is because the movie poster image server is a bit slow. Hopefully this speed issue will go away when the image proxy is deployed.
- The transforms' `isValid` method is pushing simplicity a bit far, but it also provides a kind of proto-schema to help document which keys are required for each type, and writing out the full names of the keys aids grepping.

The one thing that still bugs me is the MAGIC constant; I need to spend some time to make sure the 'Recommended' text is pushed far enough to the right. Suggestions welcome.

### DIFF
--- a/lib/movie-transform.js
+++ b/lib/movie-transform.js
@@ -1,0 +1,69 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const EXPORTED_SYMBOLS = ['movieTransform'];
+
+function MovieTransform() {
+  /*
+  Class that normalizes movie API responses into a data object containing all
+  required keys. Used by the Recommendation to generate a uniform data object
+  to pass into the RecommendationRow for rendering into the XUL DOM.
+
+  This class is stateless, so a singleton instance is created and exported.
+  */
+}
+
+MovieTransform.prototype = {
+  transform: function(data) {
+    if (!this.isValid(data)) {
+      return;
+    }
+    // NOTE: using the full keys repeatedly to aid grepping.
+    const result = {
+      type: 'movie',
+      title: `${data.enhancements.movie.title} (${data.enhancements.movie.year})`,
+      url: data.enhancements.movie.imdb_url,
+      favicon: {
+        url: data.enhancements.favicon.url,
+        color: data.enhancements.favicon.color
+      },
+      // The movie poster's dimensions are missing (recommendation-server
+      // issue #144). For now, the IMDB ratio seems to be 3:2, so just roll with that.
+      keyImage: {
+        url: data.enhancements.movie.poster,
+        height: 96,
+        width: 64
+      },
+      details: this.generateDetails(data),
+      description: data.enhancements.movie.plot
+    };
+    return result;
+  },
+  generateDetails: function(data) {
+    let items = [
+      'IMDB Rating ' + data.enhancements.movie.rating.imdb.raw,
+      data.enhancements.movie.runtime,
+      data.enhancements.movie.genre];
+    return items.join(' - ');
+  },
+  isValid: function(data) {
+    let valid = false;
+    // Just try to access every required key. If one is missing, the thrown Error
+    // will tell us it's an invalid data object. For optional keys, just check
+    // that the parent key exists (if no other check touches the parent).
+    try {
+      valid = data.enhancements.movie.title &&
+              data.enhancements.movie.imdb_url &&
+              data.enhancements.favicon.url &&
+              data.enhancements.favicon.color &&
+              data.enhancements.movie.poster &&
+              data.enhancements.movie.rating.imdb &&
+              data.enhancements.movie.runtime &&
+              data.enhancements.movie.genre;
+    } catch (ex) {}
+    return valid;
+  }
+};
+
+const movieTransform = new MovieTransform();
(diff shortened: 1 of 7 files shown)


lib/movie-transform.js
```diff
@@ -0,0 +1,69 @@
...
+
+MovieTransform.prototype = {
+  transform: function(data) {
+    if (!this.isValid(data)) {
+      return;
+    }
+    // NOTE: using the full keys repeatedly to aid grepping.
+    const result = {
+      type: 'movie',
+      title: data.enhancements.movie.title + ' (' + data.enhancements.movie.year + ')',
```
chuckharmston: Seems like an excellent candidate for a template string.

lib/movie-transform.js
```diff
@@ -0,0 +1,69 @@
...
+    // that the parent key exists (if no other check touches the parent).
+    try {
+      valid = data.enhancements.movie.title &&
+              data.enhancements.movie.imdb_url &&
+              data.enhancements.favicon.url &&
+              data.enhancements.favicon.color &&
+              data.enhancements.movie.poster &&
+              data.enhancements.movie.rating.imdb &&
+              data.enhancements.movie.runtime &&
+              data.enhancements.movie.genre;
```
chuckharmston, line 63: I'm not sure how I feel about this approach. Is there any reason we can't be more fault-tolerant on the data side and just leave out movie, runtime or genre if they're missing?

lib/ui/recommendation.js
```diff
@@ -155,52 +160,92 @@ Recommendation.prototype = {
...
       this.events.publish('recommendation-shown', this);
+      this.win.console.log('show found a dupe in the DOM, bailing.');
       return;
     }
 
-    this.prev = data;
-    const recommendation = this.row = new this.RecommendationRow(data, {
+    this.data = data;
+
+    this.win.console.log('not a dupe. this.row.url is',this.row && this.row.url,', and data.url is ',data.url);
```
chuckharmston: Unnecessary log

lib/ui/recommendation.js
```diff
@@ -155,52 +160,92 @@ Recommendation.prototype = {
...
+      type = 'movie';
+    } else {
+      type = 'wikipedia';
+    }
+    return type;
+  },
+  // Hide the recommendation after a timeout.
+  // Hide immediately if isImmediate is `true`.
+  hide: function(isImmediate) {
+    this.win.console.log('hide: ', Date.now());
```
chuckharmston: Bunch of console statements here.

chuckharmston:
A few minor notes on the code, but it otherwise looks good.

One philosophical nit: I think we should strive to be as fault-tolerant as we can. Of course we can't show a movie recommendation without an imdb ID, but we are getting data from external sources whose reliability we can't speak to.

I have a series of meetings this afternoon and won't have time to install and test it until tomorrow morning, but you're welcome to merge if you'd rather not wait that long.

lib/movie-transform.js
jaredhirsch, line 63: Yeah, the intent was to leave out validation for optional fields. I'm not clear on what fields might be missing for a given response, though; could you document the different response types in the server API docs? We're short on time, so a really simple schema should work: a list of keys, datatypes, and any constraints, like `movie.title (required): string < 255 chars`

lib/ui/recommendation.js
jaredhirsch: Thanks, yeah, I'll delete all of those

jaredhirsch:
> One philosophical nit: I think we should strive to be as fault-tolerant as we can.

I agree. Ideally, the mocks / acceptance criteria would detail the minimal required keys. I'll probably land this with the current set of required keys, and I can do a fast-follow tonight / early tomorrow to remove the optional keys from the `isValid` checks once those are documented.

jaredhirsch:
Also, thanks for the review :-)

lib/movie-transform.js
jaredhirsch: :+1:

lib/universal-search.js
```diff
@@ -27,22 +29,88 @@ function Search() {
...
+      // This looks strange, but it's required to send over the test ID.
+      const subject = {
+        wrappedJSObject: {
+          observersModuleSubjectWrapper: true,
+          object: 'universal-search@mozilla.com'
+        }
+      };
+      Services.obs.notifyObservers(subject, 'testpilot::register-variants', JSON.stringify(tests));
+
+      setTimeout(sendPing, PING_DELAY);
```
chuckharmston: We probably want to store this timeout, and clear it when they're received.

jaredhirsch: sure

PR #268, opened 2016-07-21
Remove all calls to Cu.unload to fix #267

### LINKED ISSUE #267: You're breaking Firefox by unloading Console.jsm
Please please please don't ever call `Cu.unload`. It's a bad idea, and it's led to multiple reports of brokenness (e.g. form submission inside Firefox breaks). You're calling it with a scope object - that's not how unload works: it just unloads all instances of the module everywhere (the instances are shared anyway). So the world breaks.

gijsk:
Remove all the broken `Cu.unload` calls. `Cu.unload` does not take a second argument, and so these calls don't do what their authors expect. Instead, they break all previously loaded instances. So if we unload scripts from one window, we're effectively unloading them everywhere, breaking the other consumers. That is never a good idea, and so unloading the scripts is not useful, so let's just remove them.

The only exception could be when unloading an add-on to update it, but judging by the function naming that's not the case we're dealing with here.

### DIFF
--- a/lib/universal-search.js
+++ b/lib/universal-search.js
@@ -66,17 +66,12 @@ Search.prototype = {
     // to be defined, while _stopApp deletes `win.universalSearch`.
     this._restoreSearchBar(win);
     this._stopApp(win);
-    this._unloadScripts(win);
     this._unloadStyleSheets(win);
 
     // Delete any remaining references.
     delete win.universalSearch;
 
     console.log('unloadFromWindow finish');
-
-    // We invoke console.log above, so we cannot safely unload Console.jsm
-    // until this point.
-    Cu.unload('resource://gre/modules/Console.jsm', win.universalSearch);
   },
 
   _initializePrefs: function() {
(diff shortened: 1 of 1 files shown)


jaredhirsch:
@gijsk Hi, could you please explain what this code does in the pull request description?

gijsk:
I thought the description in #267 was pretty clear, but sure...

jaredhirsch:
@gijsk Yeah, the description in #267 is quite clear. Thanks very much for opening this PR, owe you a beer in Hawaii. Merging.

jaredhirsch:
> The only exception could be when unloading an add-on to update it, but judging by the function naming that's not the case we're dealing with here.

Ah, we do need to unload the app code to upgrade. I should have tested this patch before merging. Cleanest fix is probably to revert this and just remove the Cu.unload that removes Console.jsm.

gijsk:
Wait, so do I understand correctly that the `unloadFromWindow` method is only called when the add-on is unloaded for upgrades? In that case, it would still be sensible to reorganize it and only run `Cu.unload` calls once, without their second parameter, ie for every script URL you can just run:

```
Cu.unload(SOME_URL);
```

once, instead of once per window. The second argument is currently simply ignored (or maybe it throws exceptions on the second window because the module won't be loaded anymore at that point, I'm not sure).

gijsk:
Ugh, I suck at github - @6a68 ^^ please see the previous comment.

jaredhirsch:
@gijsk lol, no worries. Thanks for your comments. 

In this addon, `unloadFromWindow` is called when the add-on code is unloaded, which indeed happens when the addon is upgraded or uninstalled. (In case you're curious, the basic flow is: [bootstrap's `shutdown` method](https://github.com/mozilla/universal-search/blob/master/bootstrap.js#L33) calls [`UniversalSearch.unload`](https://github.com/mozilla/universal-search/blob/master/lib/universal-search.js#L43), which calls [`WindowWatcher.stop`](https://github.com/mozilla/universal-search/blob/master/lib/universal-search.js#L43). That lib iterates over windows, calling [`unloadFromWindow`](https://github.com/mozilla/universal-search/blob/master/lib/universal-search.js#L62) on each. Because this add-on modifies the XBL bindings for the autocomplete popup, the shutdown process is a little bit involved; we really do need to loop over windows and remove `-moz-binding` stylesheets and so on.)

It is really unfortunate that there is such dangerous asymmetry between `Cu.import`, which loads a module into a given context, and `Cu.unload`, which unloads modules globally. I guess it's a bit late in the history of Gecko to change the `Cu.unload` function signature ;-), so I'll file a followup bug to only `Cu.unload` once.

PR #273, opened 2016-07-22
Revert "Remove all calls to Cu.unload to fix #267"

### LINKED ISSUE #267: You're breaking Firefox by unloading Console.jsm

jaredhirsch:
Reverts mozilla/universal-search#268

We still need to unload our code to upgrade successfully, but should remove the line that unloads Console.jsm globally

### DIFF
--- a/lib/universal-search.js
+++ b/lib/universal-search.js
@@ -66,12 +66,17 @@ Search.prototype = {
     // to be defined, while _stopApp deletes `win.universalSearch`.
     this._restoreSearchBar(win);
     this._stopApp(win);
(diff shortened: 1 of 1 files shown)


jaredhirsch:
Merging myself just to speed things along

PR #278, opened 2016-07-27
Use A/B testing framework to show movie cards to half of users

jaredhirsch:
@chuckharmston R?

Summary of changes:
- fetch variant data before initializing the app
- pass the data into the ui/recommendation
- use variant data to toggle whether movie cards are displayed; see `getResultType` in `lib/ui/recommendation.js` for the very simple implementation details

Still need to verify that this works with the latest Test Pilot add-on.


chuckharmston:
I couldn't figure out how Jared got this to work, so I'm closing this in favor of a commit on #316.

PR #282, opened 2016-08-02
Revert "200 movie card"

jaredhirsch:
Reverts mozilla/universal-search#261

Taking a pause on future feature development without blocking our ability to do bugfix releases


jaredhirsch:
we'll need to cherry-pick and re-land the Cu.unload bug fix commit if this is reverted

Issue #304, opened 2016-08-26
Re-enable one-off search buttons


jaredhirsch:
The fix for #286 disabled the one-off search buttons. Let's re-enable them.

_Note: there be :dragon:s here_

6a68:
oops! dupe of #294

PR #305, opened 2016-08-27
Disable one-off search buttons and simplify popup initialization

### LINKED ISSUE #286: Unable to browse the web with Universal Search enabled and latest Nightly
_Edited by @6a68: I've broken out some child bugs for individual smaller issues. Crossing them off here as I go, so interested parties can follow progress at a glance._
- [x] #292 
- [x] #293 
- [x] #294 
- [x] #295 
- [x] #296
- [x] #298 

Original bug description follows:

---

Upgraded nightly (linux64, ubuntu 15.10) to latest version (buildid 20160804030441) resulted in address bar not loading anything at all.

STR:
1. Type URL in address bar
2. Press enter or click the load button

Expected:
 Website starts to load

Actual:
 Nothing happens.

I had to disable Universal Search addon to get my nightly back into a working state.

### LINKED ISSUE #292: Key focus broken on Nightly
The latest Nightly changes have broken key focus: if the recommendation is visible, the focus is not released by the existing results when we set `selectedIndex = -1`.

Need to figure out if it's still possible to steal key focus, and if so, how.

Example:

### LINKED ISSUE #293: Mouse focus broken on Nightly
STR:
- type a single char, like 'f', in a fresh profile (so that the facebook.com result won't be hidden as a dupe)
- mouse over the regular results
- a recommendation should be visible; mouse up to it from the results list
- the regular results don't give up the blue highlight, there awkwardly seem to be two highlights:

I suspect figuring out how to remove the highlight from the results list will fix both this and the keyboard focus bug (#293).

### LINKED ISSUE #295: Nightly breakage: stop using popupshowing event (maxRows error)
The autocomplete popup's maxRows value is getting set to -1. This causes breakage described at length below.

The problem is that the popup sets maxRows to -1 when the popup closes, so that it can determine, in the popupshowing handler, whether the popup is opening due to user input (or something like that). Unfortunately, when we add a popupshowing handler inside lib/ui/popup.js, we clobber the existing popupshowing handler (XBL handlers are not executed if JS handlers for the same event are attached to the same element).

Need to remove the popupshowing handler, or replace with a popupshown handler. Same basic issue as #164.

More details and background below.

---

`maxRows` determines how many rows will be shown in the autocomplete popup. This number is set by the input to which the popup belongs. In our case, the urlbar has a `maxrows` attribute, which is normally set to 10.

On Nightly, sometimes the popup fails to render after it's opened and closed once. When this bug occurs, this error appears lots of times in the console:  `TypeError: rows[(numRows - 1)] is undefined`.

I paused the debugger inside the affected code path when the bug was happening, and it looks like the value of numRows was -1. Since the adjustHeight code assumes numRows >= 0, it naively checks `if (numRows)` before doing some height calcs. Unfortunately, in JavaScript, 0 is the only falsy number, so if numRows is -1, it tries to calculate height by counting the number of rows; when there are no rows, an error is thrown. This all happens inside the adjustHeight method:

Here's where `if (numRows)` is checked:
https://dxr.mozilla.org/mozilla-central/source/toolkit/content/widgets/autocomplete.xml#1164

Here, a few lines down, is where the error is thrown:
https://dxr.mozilla.org/mozilla-central/source/toolkit/content/widgets/autocomplete.xml#1185

### LINKED ISSUE #298: Nightly error: `property handleEvent is not callable`
Opening an error to track this issue, which seems to be triggered both by the popup opening and closing:

jaredhirsch:
@chuckharmston, @0c0w3, mind taking a look? Commit message continues below:

This commit contains a small cluster of changes that jointly fix #286:

Connecting the urlbar to the popup in an XBL constructor works far
better than the old onFirstPopupOpen approach. This is a better way of
fixing #138, and it resolves the 'maxRows - 1 is undefined' error.

Disabling one-off search by disabling the pref and removing the one-off
search buttons from the XUL DOM (via hidden=true) closes #292,
closes #293, closes #298--but those bugs aren't really fixed, just
hidden. Filed #304 to re-enable one-off search as time allows. Note that
the disabletab attribute was added just to make the set of attributes in
our one-off-search-buttons element match the element in FF.)

Switching our popup.js listener from the 'popupshowing' event to the
'popupshown' event fixes #295, with some cascading changes related to
renaming the function and the signal fired by the function.

\o/

### DIFF
--- a/chrome/content/binding.xml
+++ b/chrome/content/binding.xml
@@ -19,11 +19,18 @@
         <children/>
         <xul:vbox anonid="one-off-search-buttons"
                   class="search-one-offs"
+                  hidden="true"
                   compact="true"
                   includecurrentengine="true"
+                  disabletab="true"
                   flex="1"/>
       </xul:hbox>
     </content>
+    <implementation>
+      <constructor><![CDATA[
+        this.mInput = window.gURLBar;
+      ]]></constructor>
+    </implementation>
   </binding>
 
   <![CDATA[
--- a/lib/metrics.js
+++ b/lib/metrics.js
@@ -26,8 +26,8 @@ function Metrics(opts) {
 
 Metrics.prototype = {
   init: function() {
-    // Start listening when the popup is about to open.
-    this.events.subscribe('before-popup-show', this.interactionStart);
+    // Start listening when the popup has just opened.
+    this.events.subscribe('after-popup-shown', this.interactionStart);
 
     // Detect when a navigation event occurs.
     this.events.subscribe('before-keyboard-navigate', this.onKeyNavigate);
@@ -40,7 +40,7 @@ Metrics.prototype = {
     this.events.subscribe('after-popup-hide', this.interactionEnd);
   },
   destroy: function() {
-    this.events.unsubscribe('before-popup-show', this.interactionStart);
+    this.events.unsubscribe('after-popup-shown', this.interactionStart);
     this.events.unsubscribe('before-keyboard-navigate', this.onKeyNavigate);
     this.events.unsubscribe('before-click-navigate', this.onClickNavigate);
     this.events.unsubscribe('urlbar-change', this.interactionStart);
--- a/lib/ui/popup.js
+++ b/lib/ui/popup.js
@@ -15,9 +15,8 @@ function Popup(opts) {
   this.popupOpenObserver = null;
 
   this.afterPopupHide = this.afterPopupHide.bind(this);
-  this.beforePopupShow = this.beforePopupShow.bind(this);
+  this.afterPopupShown = this.afterPopupShown.bind(this);
   this.onResultsMouseMove = this.onResultsMouseMove.bind(this);
-  this.onFirstPopupOpen = this.onFirstPopupOpen.bind(this);
   this.onClick = this.onClick.bind(this);
 }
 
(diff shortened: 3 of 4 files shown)


0c0w3:
Looks straightforward to me... I still can't figure out why/how mInput becomes falsey. It's getting set in urlbarBinding's version of _openAutocompletePopup but then it's falsey when autocomplete.xml accesses it a moment later. But anyway, glad this fixes it.

jaredhirsch:
> I still can't figure out why/how mInput becomes falsey.

@0c0w3 Thanks for taking a look! I gotta say, it's a relief to know that you're mystified too ;-P

chuckharmston:
I'm not super thrilled that the solution is to hide the one-off search buttons (mostly since we're kind of smothering the other place you can make one-off searches), but I trust your judgement that it would be much more challenging to try to work with it.

Code looks good. I probably won't have a chance to test the code on my machine this week, so if you're comfortable, feel free to merge it and ship a new releases.

jaredhirsch:
> I'm not super thrilled that the solution is to hide the one-off search buttons

Yup, I agree. Filed #304 to try to re-enable them after Nightly is fixed, can work on that this coming sprint maybe.

jaredhirsch:
still works for me in today's Nightly, merging

PR #310, opened 2016-09-14
Mark addon as multiprocess compatible (closes #291)

### LINKED ISSUE #291: Universal Search should be marked as multiprocess compatible
Currently Universal Search isn't marked as multiprocess compatible. Although it seems to work, there may still be some incompatibilities that are masked by Firefox's shims.

The shims may also lead to performance issues, although I doubt Universal Search would really be hitting those that hard: https://developer.mozilla.org/Add-ons/SDK/Guides/Multiprocess_Firefox_and_the_SDK#Testing_with_shims_enabled

Setting the flag for Universal Search should be as simple as adding the following to `install.rdf`:

```
<em:multiprocessCompatible>true</em:multiprocessCompatible>
```

Obviously, it should be tested to make sure it does work.

chuckharmston:


jaredhirsch:
lgtm and everything seems to work fine with the shims removed. merging :beers:

PR #323, opened 2016-10-17
Update urlbar state management to fix Enter key in Nightly

### LINKED ISSUE #322: Can't select entries in the address bar's drop-down list in Nightly 52
STR:
1. Enable the Universal Search add-on version 1.0.11.
2. Enter some text in the address bar.
3. Use the arrow keys to select a page in address bar's drop-down list.

RESULT:
The selected page is not loaded. Instead, the entered text is passed to my default search provider. You _can_ select a page from the drop-down list using the mouse.

This is a regression from [Firefox bug 1306639](https://bugzilla.mozilla.org/show_bug.cgi?id=1306639) in Firefox Nightly 52. Disabling the Universal Search add-on fixes the problem.

jaredhirsch:
- Upstream bug 1306639 changed how the urlbar state is set. Update how
  we set urlbar state to work with the new approach.

Fixes #322.

@chuckharmston mind giving this a try? seems to fix the issue for me in nightly, but also keeps things working as usual in dev edition.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 1.0.12
+
+**Date:** 2016/10/17
+
+**Download:** [`universal-search-1.0.12.xpi`](https://s3-us-west-2.amazonaws.com/universal-search/universal-search-1.0.12.xpi)
+
+**Summary:** Fixes broken Enter key in Nightly ([issue 322](https://github.com/mozilla/universal-search/issues/322)).
+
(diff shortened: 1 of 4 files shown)


6a68:
Please don't merge this quite yet. I want to roll a version update into this PR.

chuckharmston:
Verified that this patch works as expected in each channel.

chuckharmston:
1.0.12 released and verified.

PR #326, opened 2016-10-24
Change the update URL

clouserw:
This is a new update URL.  I'm not changing the one in update.rdf because that file would no longer be in use.  If you r+ this patch, next steps:

1) Make a new add-on version with this URL in it
2) Send new .xpi to clouserw to put on S3
3) Push new add-on version live, people will upgrade it and change their updateURL to s3 (which will have the new version, so it won't upgrade again).
4) Patch testpilot to point to https://testpilot.firefox.com/files/universalsearch/latest for the install links


jaredhirsch:
@clouserw @chuckharmston what's the status of this PR? I'm guessing we do not need it?

clouserw:
I reckon

